PR #3, opened 2016-11-08
Fix FreeBSD build

myfreeweb:
(Maybe there's a better cross-platform way to use the `bswap` functions...?)


haberman:
Hmm, I think it might make sense to move all the endianness code into a separate header `endian.h` to keep all the ifdefs out of `bloaty.h`.

haberman:
I suppose that a refactoring to separate out endianness into a separate file could wait until later.

PR #6, opened 2016-11-08
Add a using declaration for functions used from cmath.

safarmer:
Add using directives for the functions used in <cmath>

This was causing compilation to fail with G++ and Clang++ on Ubuntu 16.0.4.


haberman:
Thanks for the PR! I just merged a PR that explicitly qualifies these calls: https://github.com/google/bloaty/pull/5

PR #7, opened 2016-11-08
Use $(MAKE) to invoke sub-makes

emaste:
On non-GNU systems GNU make might be installed as 'gmake'.


myfreeweb:
I've included that change in #3 already… do the `__bswap_16` endian things not error for you?

emaste:
Ah indeed, I didn't see your existing #3 pull request. It seems my dev system had a hack to add the Linux-style definitions from some time ago; the build does fail on an unmodified host.

The FreeBSD base system c++filt also needs a per-line fflush added in r308465
https://svnweb.freebsd.org/changeset/base/308465

PR #11, opened 2016-11-09
Update submodules in serial

brookst:
The submodule update applies to all targets, so don't run an update per
target. Use a dummy pattern match so that make knows these are related.

This prevents the make failing when first run as e.g. `make -j2`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,12 @@ src/main.o: src/main.cc src/bloaty.h $(RE2_H)
 third_party/re2/obj/libre2.a: third_party/re2/Makefile
 	$(MAKE) -C third_party/re2 CPPFLAGS="-ffunction-sections -fdata-sections -g"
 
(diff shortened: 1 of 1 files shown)


haberman:
Can you enlighten me about what the `%` are for and why you used them instead of `_`?

brookst:
It's a bit of a hack. The problem being solved is that make runs the submodule init/update step once for each target. If running in parallel this causes all bar one git instance to error out due to a file lock, in turn causing make to error out.

The hack is to convince make that the rule should only run once. This relies on pattern matching - the `%` character is a wildcard I expect to only match `_`. The rule will then only run for that single match, expecting all targets to be updated. There's a stack overflow answer that explains this (and a non-portable alternative) here: http://stackoverflow.com/a/3077254/1780018

The touch just limits how often the update will happen without a change to `.gitmodules`

haberman:
Thanks for the explanation. Could you include a link to the SO thread for the benefit of future readers?

brookst:
Sure, added in a comment.

Issue #10, opened 2016-11-09
segfault


twiz718:
build from 83fe66b2386d54840776c965467070c26f695b1a

```
./bloaty bloaty -d             
bloaty: option '-d' requires an argument
[1]    1316 segmentation fault  ./bloaty bloaty -d

uname -a
Darwin akhanin.local 16.1.0 Darwin Kernel Version 16.1.0: Thu Oct 13 21:26:57 PDT 2016; root:xnu-3789.21.3~60/RELEASE_X86_64 x86_64

otool -L bloaty 
bloaty:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
```

haberman:
This was fixed in https://github.com/google/bloaty/pull/14

PR #14, opened 2016-11-09
Fixes issue #10

Ethiraric:
Segfault when an option is not provided an argument:
    bloaty ./bloaty -d


Ethiraric:
I signed it!

haberman:
Thanks for the fix!

PR #29, opened 2016-11-14
Add `#define __STDC_LIMIT_MACROS` to bloaty.h

jchl:
This prevents a compile error due to an undefined symbol UINT64_MAX
on platforms where stdint.h doesn't otherwise define it.

### DIFF
--- a/src/bloaty.h
+++ b/src/bloaty.h
@@ -20,6 +20,7 @@
 #define BLOATY_H_
 
 #include <stdlib.h>
+#define __STDC_LIMIT_MACROS
 #include <stdint.h>
 #ifdef __FreeBSD__
 #include <sys/endian.h>


jchl:
I signed it!

src/bloaty.h
```diff
@@ -20,6 +20,7 @@
 #define BLOATY_H_
 
 #include <stdlib.h>
+#define __STDC_LIMIT_MACROS
```
haberman: I wonder if it would be slightly safer to put this before *all* includes. For example, this would fail if stdlib.h included stdint.h, because by this point it would be too late to `#define __STDC_LIMIT_MACROS`.

Also a comment mentioning why we need it would be helpful. Something like:

```
#define __STDC_LIMIT_MACROS  // To make stdint.h define UINT64_MAX
```

haberman:
Actually don't worry about it for now. We can cross that bridge when we come to it.

PR #31, opened 2016-11-16
bloaty: Fix regexp name munging.

pchx:
Note: Regexp name munging happens before C++ demangling, so collapsing
symbols by namespace (s/re2::.*/RE2/) doesn't quite work as you might expect.


haberman:
Yes the regex munging needs some work. There needs to be a more comprehensive and principled approach to this. But your change will at least un-break the existing functionality for now.

Issue #21, opened 2016-11-11
Long file names and symbols get cut off


jdm:
Eg.
```
jdm@nightblogger:~/bloaty$ ./bloaty -d compileunits ../servo/target/debug/servo 
     VM SIZE                                                                                        FILE SIZE
 --------------                                                                                  --------------
  90.8%   109Mi [None]                                                                             462Mi  97.7%
   5.6%  6.72Mi [Other]                                                                           6.72Mi   1.4%
   0.3%   355Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   355Ki   0.1%
   0.3%   346Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   346Ki   0.1%
   0.3%   316Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   316Ki   0.1%
   0.2%   303Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   303Ki   0.1%
   0.2%   285Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   285Ki   0.1%
   0.2%   265Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   265Ki   0.1%
   0.2%   265Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   264Ki   0.1%
   0.2%   258Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   258Ki   0.1%
   0.2%   257Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   257Ki   0.1%
   0.2%   253Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   253Ki   0.1%
   0.2%   221Ki /home/jdm/servo/target/debug/build/mozjs_sys-661d4efe7c7ca939/out/js/src/Unified   221Ki   0.0%
[...]

PR #36, opened 2016-12-13
Added -w flag for avoiding truncation.

haberman:
This can be one solution for https://github.com/google/bloaty/issues/21

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -1550,6 +1550,7 @@ Options:
   -v               Verbose output.  Dumps warnings encountered during
                    processing and full VM/file maps at the end.
                    Add more v's (-vv, -vvv) for even more.
+  -w               Wide output; don't truncate long labels.
   --help           Display this message and exit.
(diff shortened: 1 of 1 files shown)


haberman:
Want to take a look @fischman?

src/bloaty.cc
```diff
@@ -1637,6 +1637,8 @@ bool BloatyMain(int argc, char* argv[], const InputFileFactory& file_factory,
       verbose_level = 2;
     } else if (strcmp(argv[i], "-vvv") == 0) {
       verbose_level = 3;
+    } else if (strcmp(argv[i], "-w") == 0) {
+      max_label_len = 100000000;
```
fischman: Why not std::numeric_limits<size_t>::max() or SIZE_MAX or however your lowest denominator platform spells it?
((size_t)-1?)

Also, mention in help/usage string?

Otherwise LGTM.

haberman: Done.

PR #37, opened 2017-02-12
Enable more strict compiler checks

sievlev:
Added '-W' compiler flag and made some code improvements

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 
 # Disable -Wsign-compare because StringPiece currently unhelpfully defines
 # size() as ssize_t instad of size_t.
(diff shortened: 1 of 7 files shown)


sievlev:
I signed it!

haberman:
Thanks for this! I had no idea that `-W` included some things that are not in `-Wall`.

For the unused variables, I think I'd slightly prefer: `Type /* unused_name */` instead of just `Type`. That gives the reader more clue about what the unused param is for.

So for example: `const std::string& /* msg_regex */`

sievlev:
Ok, I will update my patch.

PR #38, opened 2017-03-20
CONTRIBUTING: fix CLA links

wjt:
There must be no whitespace between the `]` and `(` of a link (apparently).

I assume I will have to sign one of these CLAs before this patch is accepted…


wjt:
I signed it! 🎉

PR #41, opened 2017-03-21
Add --no-demangle option to display fully mangled C++ names.

cwgreene:
Useful for side by side comparison with other tools (that haven't been passed through c++filt).

Now with tests! (Well, at least an attempt at them).


cwgreene:
Oh well that's interesting. The demangler looks like it works differently on mac os x. Will investigate.

cwgreene:
Tests blocked by #43

haberman:
Thanks for this PR and sorry I've been so unresponsive! I've got some time to devote to Bloaty at the moment so I'm trying to take care of all the outstanding issues.

I like the idea, but I think it might be better to have it be a separate data source (maybe `ccsymbols`) instead of a command-line flag.

Might be better to leave out tests for now since OS X doesn't have `c++filt`.

haberman:
This change is obsoleted: now mangled and non-mangled symbols have different data sources, which I think is a cleaner approach overall. Thanks for helping to spur the idea!

PR #46, opened 2017-03-30
Switch order of touch args

msabramo:
so as not to create a `-r` file on OS X.

Without this if I do a fresh checkout on OS X and then`make`, I end up with a `-r` file:

```
$ ls -l -- -r
-rw-r--r--  1 abramowi  wheel  0 Mar 29 17:44 -r
```


haberman:
Thanks for the fix!

PR #50, opened 2017-04-06
src/bloaty.h: Use bswap builtins.

### LINKED ISSUE #47: bloaty build failure with musl; bloaty.h uses bswap and friends without proper include
Needs to `#include "<byteswap.h>"` on glibc and musl, can't speak for other platforms.

edit: That is to say, 'byteswap.h' is the header that should be included on glibc when using these functions, see http://man7.org/linux/man-pages/man3/bswap.3.html .  Clearly bloaty builds against glibc as-is, likely one of the headers happens to include 'byteswap.h' for now.

dtzWill:
Makes code a bit neater and the builtins
are supported by at least gcc and clang.

Testing required to confirm no breakage
introduced on OSX or FreeBSD.

Removes dependence on non-standard
libc-specific mechanism for determining
how to efficiently swap bytes on a given arch:
instead, trust the compiler! :)

Fixes #47.

### DIFF
--- a/src/bloaty.h
+++ b/src/bloaty.h
@@ -22,9 +22,6 @@
 #include <stdlib.h>
 #define __STDC_LIMIT_MACROS
 #include <stdint.h>
-#ifdef __FreeBSD__
-#include <sys/endian.h>
-#endif
 
 #include <memory>
 #include <set>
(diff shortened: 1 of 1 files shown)


dtzWill:
@googlebot CLA has been signed!

haberman:
Thanks for the PR!

I agree with your assessment of "trust the compiler." Actually I think we could take that principle even a step further than you have done here.

I recently discovered that compilers these days are smart enough to emit byte-swapping instructions with no intrinsics at all! See: https://godbolt.org/g/AOOctq

I'd love to replace the existing functions with this approach and remove all platform/compiler-specific code. Would you be up for revising your PR to take this approach?

haberman:
I went ahead and implemented the fully-portable approach myself. Thanks for providing the motivation though!

dtzWill:
Sure thing, thanks! :+1:

Will be testing to confirm re:#37 shortly, but I expect it does the trick :).

PR #51, opened 2017-04-09
Generate test cases on build reproducibly.

cwgreene:
This change will allow avoiding committing binary blobs to the repo
by ensuring that we use a well defined gcc (via docker) version for
creation of binaries from c source snippets.

The Makefile has been updated to run a sha256 check to ensure
that the built blobs are exactly as expected. When adding tests, the
.expected_shasums file will need to be updated.


haberman:
Thanks for the PR! I'd prefer to keep the binary blobs checked in though. Otherwise Docker is required to run the tests. I'd prefer for Docker only to be required when someone is modifying the test data.

Can we just change `make_test_files.sh` so that it is deterministic?

cwgreene:
Yes, we can do that; and might have to. Docker doesn't run on the mac os builds.

cwgreene:
Problem is now understood; CC had been set, so docker wasn't being invoked for compilation. AR was. Closing pull request for make_test_files.sh only change.

Issue #40, opened 2017-03-21
No good way to add test cases.


cwgreene:
make_tests_files.sh after running will result in the binaries being different, which makes it unsuitable for reproducing the test cases (and also means that new test cases cannot be verified easily; but trust me, I'm sure all the binary blobs that I will commit to this repo can be trusted ;) ). Arguably, only make_test_files.sh should be committed, and have the test cases get automatically regenerated. This has the obvious drawback of not having a canonical test cases (and requiring careful control over compiler versions).<sup>1</sup>

Ideas?

<sup>1</sup> More seriously, it basically means that at the very least, one would have to add cross compilers to the toolchain.

cwgreene:
So it looks like the chromium project has sucessfully gotten deterministic builds working on linux.

https://www.chromium.org/developers/testing/isolated-testing/deterministic-builds

One option would be to use a docker image to provide a linux build environment to generate the binaries locally.

haberman:
Sorry for the slow reply!

I like the idea of using a docker image to regenerate the test cases reliably!

I think that is a better solution than regenerating the test cases at build time. The things being asserted about the input files could easily fail to be true for different versions/configurations of the toolchain.

I'm happy to accept patches to use a stable Docker image for regenerating these.

cwgreene:
So the good news is that it's relatively straightforward to make `make_test_files.sh` compile with a dockerified gcc-4.9, and I've succesfully built binaries and shared libraries that are identical when built on both linux and mac os.

The static libraries created through 'ar' however, are presenting their own difficulties. I don't yet know the reason, but it could be due to some metadata about the filesystem. Builds on a given machine are consistently giving the same hashes, so it's unlikely that it's due to the temp directory names getting encoded. I'll poke around and see what's going on.

cwgreene:
Previous claim was in error. The hashes are different from build to build. Looks like there are timestamps in the archive.

cwgreene:
Got it. Change
```
$AR rcs
```
to 
```
$AR Drcs
```
For deterministic archiving purposes.

haberman:
After thinking about this a lot, I don't think we should require reproducible builds for test cases. It's too hard, and I don't think it's truly necessary.

[...]

PR #52, opened 2017-04-09
Use docker for consistent Test Case Generation.

cwgreene:


haberman:
Closing for the reasons mentioned in https://github.com/google/bloaty/issues/40.

PR #54, opened 2017-04-20
Removed platform-specific bswap() function calls.

haberman:
It turns out compilers these days are smart enough
to recognize when generic byte-swapping code can
be substituted with a bswap() instruction.  So this
change removes the platform-specific includes and
function calls without sacrificing any performance.

This should fix https://github.com/google/bloaty/issues/47, though I can't verify for sure because I can't figure out how to build with musl and a C++ standard library.

### DIFF
--- a/src/bloaty.h
+++ b/src/bloaty.h
@@ -22,9 +22,6 @@
 #include <stdlib.h>
 #define __STDC_LIMIT_MACROS
 #include <stdint.h>
-#ifdef __FreeBSD__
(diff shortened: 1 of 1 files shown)


dtzWill:
I can confirm 2fdc04e957cacd558dd63b9f997a24d1e3990ded  fixes #47 for me! Can be closed when this is merged, thanks!

haberman:
Great, thanks!

PR #55, opened 2017-04-20
Warn when we can't read compressed debug info.

haberman:
This just improves the error message slightly.


EricRahm:
Looks like we added support in #230 and #233, lets close this out.

PR #57, opened 2017-10-04
Added a CMakeLists.txt, and updated re2/googletest.

haberman:


CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+
+cmake_minimum_required (VERSION 2.6)
+project (Bloaty)
+
+set(RE2_BUILD_TESTING OFF CACHE BOOL "enable testing for RE2" FORCE)
+add_subdirectory(third_party/re2)
+include_directories(third_party/re2)
+
+include_directories(.)
```
xfxyjwf: include_directories(src)?

CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+
```
xfxyjwf: Remove this empty line?

haberman: Done.

CMakeLists.txt
haberman: Doesn't seem to be required? It is compiling without this.

xfxyjwf: Hmm. I saw that in the source code it's doing 'include "bloaty.h"' so I assume you need to put src in the include path. Is it there by default?

PR #75, opened 2017-11-07
Rewrote arg parsing to be clearer and more C++-like.

haberman:


src/bloaty.cc
```diff
@@ -1657,93 +1657,138 @@ USAGE: bloaty [options] file... [-- base_file...]
...
         options->set_sort_by(Options::SORTBY_BOTH);
       } else {
-        THROWF("unknown value for -s: $0", argv[i]);
+        THROWF("unknown value for -s: $0", option);
       }
-    } else if (strcmp(argv[i], "-v") == 0) {
+    } else if (args.TryParseFlag("-v")) {
       options->set_verbose_level(1);
-    } else if (strcmp(argv[i], "-vv") == 0) {
+    } else if (args.TryParseFlag("-v")) {
```
crujose: This should have been -vv

haberman: Good catch, thanks! Fixed.

Issue #43, opened 2017-03-22
Demangling of C++ names of Linux Binaries does not work on Mac OS


cwgreene:
Steps to reproduce:

Compare output of `tests/testdata/linux-x86_64/oldbloaty.bin` on Linux and Mac.

Linux:
```
     VM SIZE                                                                                        FILE SIZE
 --------------                                                                                  --------------
  20.0%  99.0Ki [Unmapped]                                                                        6.55Mi  94.4%
  62.1%   307Ki [Other]                                                                            308Ki   4.3%
   2.5%  12.3Ki re2::RE2::Match(re2::StringPiece const&, int, int, re2::RE2::Anchor, re2::String  12.3Ki   0.2%
   1.6%  7.83Ki re2::unicode_groups                                                               7.83Ki   0.1%
   1.5%  7.56Ki re2::NFA::Search                                                                  7.56Ki   0.1%
   1.2%  5.76Ki re2::BitState::TrySearch                                                          5.76Ki   0.1%
   1.0%  4.81Ki _dwarf_exec_frame_instr                                                           4.81Ki   0.1%
   0.9%  4.49Ki re2::DFA::DFA                                                                     4.49Ki   0.1%
   0.9%  4.34Ki re2::Regexp::Parse                                                                4.34Ki   0.1%
   0.8%  4.20Ki re2::RE2::Init                                                                    4.20Ki   0.1%
   0.8%  4.09Ki re2::Prog::IsOnePass                                                              4.09Ki   0.1%
   0.8%  4.04Ki re2::Compiler::PostVisit                                                          4.04Ki   0.1%
   0.8%  4.01Ki _dwarf_internal_srclines                                                          4.01Ki   0.1%
   0.8%  3.91Ki re2::Regexp::FactorAlternationRecursive                                           3.91Ki   0.1%
   0.8%  3.86Ki bool bloaty::(anonymous namespace)::ForEachElf<bloaty::(anonymous namespace)::Do  3.86Ki   0.1%
   0.8%  3.77Ki re2::DFA::RunStateOnByte                                                          3.77Ki   0.1%
   0.7%  3.68Ki re2::unicode_casefold                                                             3.68Ki   0.1%
   0.7%  3.40Ki re2::DFA::InlinedSearchLoop(re2::DFA::SearchParams*, bool, bool, bool) [clone .c  3.40Ki   0.0%
[...]

PR #78, opened 2017-11-09
Expanded and refined demangling support.

haberman:


haberman:
Fixes #43, since we strip leading underscores ourselves now for mach-o.

PR #86, opened 2017-11-18
Generalized ElfFile::StructReader to StructPtr.

haberman:


haberman:
I think this is the wrong approach going forward.

Issue #23, opened 2016-11-13
Support for stripped files?


shooshx:
My binary size is about 30MB
it is stripped from debug information
the unstripped binary is 250MB
I want to be able to use bloaty on my stripped binary but using the debug information from the unstripped file (this is something any debugger can do)

Also, for Mach-O, Xcode regularly produces stripped files and a dSYM files besides them that contains the symbols information

haberman:
I agree this needs to happen for Mach-O since OS X always puts debug symbols separate from the binary.

Supporting this on other systems is an interesting idea. It would get tricky if you had more than one input file -- you'd need a way of specifying which debug info file goes with which input file. But as a start I could see adding a command-line switch for specifying a file that all the symbols and debug info should come from.

haberman:
Oh, looking at this a bit more it seems like there are established conventions for how to find debugging symbols given an executable:

https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html

It seems like bloaty could implement these conventions so that no extra command-line arguments would be required.

shooshx:
that's just gdb, clang and Xcode do other things and while this convention can be a good default, I would still prefer to have  full control using a command line argument

DanAlbert:
+1 for an option other than the /usr/lib/debug convention. Relying on that convention wouldn't work for Android.

haberman:
I'm starting work on implementing this.

We can always confirm that a debug file matches a binary by matching the build id. So the question is: what is the best way to find the debug info file?

I'm thinking of implementing this like so:

1. a command-line flag like `--debug-file=<FILENAME>` will tell Bloaty to use this file for debug info, but not add its sizes to the overall profile. We'll try to match its build id to one of the input files. If none of the input files match, either error or warn.

2. if someone requests it, I could implement something that looks for the debug file by using the gnu debuglink and/or /usr/lib/debug.

haberman:
Please give https://github.com/google/bloaty/pull/94 a try if you get a chance and see if it works for you.

haberman:
Fixed in https://github.com/google/bloaty/pull/94.

jchl:
Support for `.gnu_debuglink` sections would be really useful, especially when running bloaty on a large number of files simultaneously.

haberman:
@jchl Feel free to open a new issue for this.

PR #94, opened 2017-11-24
Added support for separate debug files.

haberman:


haberman:
Fixes https://github.com/google/bloaty/issues/23

PR #118, opened 2018-05-02
added install targets for bloaty binary and library

psteinb:
when `bloaty` is built and I invoke `make install` neither the binary nor the library is put into the folder referenced by `$CMAKE_INSTALL_PREFIX`. this PR adds this feature to the cmake infrastructure. 
I am yet unable to rid the install tree of all the `third_party` dependencies.


psteinb:
I signed it!

PR #119, opened 2018-05-02
added install statements for bloaty app and library + header

psteinb:
see #118 for details. I am recreating this PR to make the google bot happy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,18 +105,8 @@ add_library(libbloaty
     third_party/demumble/third_party/libcxxabi/cxa_demangle.cpp
     )
 
-  set_target_properties(libbloaty PROPERTIES
-    OUTPUT_NAME bloaty
-    PUBLIC_HEADER src/bloaty.h
-    )
 
 set(LIBBLOATY_LIBS libbloaty libprotoc re2 capstone-static)
-install(TARGETS libbloaty
-  EXPORT ${PROJECT_NAME}Targets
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -105,7 +105,18 @@ add_library(libbloaty
     third_party/demumble/third_party/libcxxabi/cxa_demangle.cpp
     )
 
+  set_target_properties(libbloaty PROPERTIES
+    OUTPUT_NAME bloaty
+    PUBLIC_HEADER src/bloaty.h
```
haberman: I don't want to install this, at least for now. I don't consider the C++ interfaces public or stable: https://github.com/google/bloaty/blob/master/src/bloaty.h#L15

CMakeLists.txt
```diff
@@ -105,7 +105,18 @@ add_library(libbloaty
...
+    )
+
 set(LIBBLOATY_LIBS libbloaty libprotoc re2 capstone-static)
+install(TARGETS libbloaty
```
haberman: Ditto, I don't want to install the library, only the statically-linked binary.

psteinb:
thanks, didn't see that statement in the header file. would be wonderful if this would be documented outside the source code somewhere. 

I left the `install(EXPORT` statement there as it would help downstream consumers of bloaty to pick up the binary as long as they add the prefix install path to `CMAKE_PREFIX_PATH`

PR #129, opened 2018-07-10
Add support for exporting to tsv, in addtion to csv

juliehockett:
Helpful when symbols can contain commas.

### DIFF
--- a/src/bloaty.h
+++ b/src/bloaty.h
@@ -442,6 +442,7 @@ struct RollupRow {
 enum class OutputFormat {
   kPrettyPrint,
   kCSV,
+  kTSV,
 };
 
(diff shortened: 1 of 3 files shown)


src/bloaty.h
```diff
@@ -500,18 +504,18 @@ struct RollupOutput {
   bool diff_mode_ = false;
 
   void PrettyPrint(size_t max_label_len, std::ostream* out) const;
-  void PrintToCSV(std::ostream* out) const;
+  void PrintToCSV(std::ostream* out, bool tabs=false) const;
```
haberman: Bloaty follows the Google C++ style guide, which generally disallows default arguments. Please use explicit parameters only.

PR #144, opened 2018-09-13
Suppress warnings from protobuf

brookst:
Protobuf generated code has many `unused parameter` warnings with GCC 8.
Telling CMake protobuf is a system library ignores these spurious warnings.

Additionally, I opened  aquynh/capstone#1247 to clear up a strncpy warning.

There's one last warning I see in [nico/demumble](//github.com/nico/demumble):

    bloaty/third_party/demumble/third_party/libcxxabi/cxa_demangle.cpp: In function ‘const char* __cxxabiv1::{anonymous}::parse_type(const char*, const char*, C&) [with C = __cxxabiv1::{anonymous}::Db]’:
    bloaty/third_party/demumble/third_party/libcxxabi/cxa_demangle.cpp:2206:25: warning: this statement may fall through [-Wimplicit-fallthrough=]
                         if (first+1 != last)
                         ^~
    bloaty/third_party/demumble/third_party/libcxxabi/cxa_demangle.cpp:2251:21: note: here
                     default:
                     ^~~~~~~

That file is 2 years out of date relative to [llvm-mirror/libcxxabi/src/cxa_demangle.cpp](//github.com/llvm-mirror/libcxxabi/blob/master/src/cxa_demangle.cpp). I'm not sure if that could be pulled in as-is or if dependencies have been introduced into it - I'll take a look some other time.


haberman:
I think this is a good idea, sorry for the delay. If you want to rebase I'd be happy to merge.

brookst:
Ok, testing this with GCC 8.2.1 and Clang 8.0.0, I only see a `inconsistent-missing-override` message in protobuf with the Clang build. I'm not sure why the `SYSTEM` option doesn't silence that as well but at least all the `unused-parameter` spam is cleaned up from both compilers.

I added a few other clean-ups; using memmove, updating googletest and turning off a spurious warning.

Issue #142, opened 2018-08-24
Make configure option to support system packages rather than submodules


marxin:
Hello.

I really like the package but as it is it's not feasible for distribution. Can you please make it using system packages?

haberman:
I'm happy to accept a PR for this. I'm not sure exactly when I would get to it.

I want the tarball to be buildable on its own. But I would like to see CMake options that let you prefer system libs instead.

haberman:
This is fixed by https://github.com/google/bloaty/pull/155

marxin:
Thank you for that. Can you please document how can one use it? I still see submodule cloning during cmake build by default?

PR #152, opened 2018-12-06
Add an option to use system protobuf.

bertptrs:
As stated in #142, it would be nice to have an option to use system libraries rather than compiling everything for this package alone. This patch implements just that.

For this to work properly, `liblibbloaty` can no longer be a static object since it may have to depend on the shared libprotobuf instance. I hope this is not too big of an issue.

On my laptop, compilation times are pretty much halved, as is the binary size.


bertptrs:
It now no longer creates a shared library by default, since apparantly Apple doesn't like mixing shared and static libraries.

haberman:
Thanks for the pull request and sorry for the delay. I ended up merging https://github.com/google/bloaty/pull/155 which is a similar idea but larger scope.

PR #155, opened 2019-01-05
Use system deps for protobuf, capstone and re2

swills:
Note that with this system deps are used if they are found. I could add an option to only use them if requested or disabling using them if found if you like.

Also:

* add option to disable setting build id for systems where this flag is not supported by ld
* add option to disable installing cmake target files in case they aren't needed

I could make these separate pull requests if you like.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,34 @@ project (Bloaty VERSION 1.0)
 # Options we define for users.
 option(BLOATY_ENABLE_ASAN "Enable address sanitizer." OFF)
 option(BLOATY_ENABLE_UBSAN "Enable undefined behavior sanitizer." OFF)
+option(BLOATY_ENABLE_CMAKETARGETS "Enable installing cmake target files." ON)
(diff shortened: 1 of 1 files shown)


swills:
Forgot to mention, this is similar in spirit to #152 but a bit of a different approach. Also, I suppose finding the system deps via non-pkg-config methods would be a good idea, but I haven't done that yet.

FWIW, I think supporting using these from the system makes sense since re2, capstone and protobuf are pretty widely packaged in various Linux distributions:

https://repology.org/metapackage/re2/versions
https://repology.org/metapackage/capstone/versions
https://repology.org/metapackage/protobuf/versions

I think having this change will make it easier to package bloaty, which isn't as widely packaged yet:

https://repology.org/metapackage/bloaty/versions

Also, it might be good to add another ci target which has these deps installed and tests with them. I can do that as part of this pull request if you like, but wanted to get feedback first.

haberman:
Sorry for the slow reply. I think this is great. We can refine this over time (with more tests, searching methods, etc). but this seems like a good start.

PR #163, opened 2019-05-20
Document dependency on the protobuf compiler

psmarshall:
I got build failures looking for protoc when trying to build from source, it seems like that's a dependency but is not documented.

Not sure if you want to suggest that people install it using apt-get or some other method, so happy if you accept this PR or go with something else that documents the dependency.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ This is not an official Google product.
 
 ## Building Bloaty
 
+To build Bloaty from source, the protobuf compiler is needed.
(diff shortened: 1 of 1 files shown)


psmarshall:
I'm actually just realising that protobuf is pulled in as a dependency so this is probably not the correct fix to my problem.

psmarshall:
Filed an issue instead: https://github.com/google/bloaty/issues/164

Issue #170, opened 2019-08-07
Unescaped commas in CSV output


chrisroman:
When running `./bloaty bloaty --csv`, I get the following output:
```
sections,vmsize,filesize
__TEXT,__const,2151456,2151456
__TEXT,__text,1321560,1321560
__DATA,__bss,1050450,0
__DATA,__data,449104,449104
String Table,325048,325048
__DATA,__const,305856,305856
Symbol Table,248208,248208
__TEXT,__cstring,95128,95128
Export Info,43544,43544
__TEXT,__gcc_except_tab,19040,19040
__TEXT,__unwind_info,11968,11968
[10 Others],11880,9762
Lazy Binding Info,9568,9568
Weak Binding Info,4784,4784
[__TEXT],4734,4630
Binding Info,3824,3824
Function Start Addresses,3304,3304
__TEXT,__eh_frame,3304,3304
[__LINKEDIT],3008,0
Rebase Info,2264,2264
Indirect Symbol Table,2240,2240
```

However, if you look at the 2nd row `__TEXT,__const,2151456,2151456`, you'll see that the section name is `__TEXT,__const` which messes up the CSV formatting. I would expect it to be something like `"__TEXT,__const",2151456,2151456`.

This is also the case when adding the `-d compileunits`.

haberman:
This is a silly error in the code. The code defines `CSVEscape()` but does not use it.

PR #174, opened 2019-09-01
Properly escape CSV.

haberman:
Fixes: https://github.com/google/bloaty/issues/170


haberman:
This should have unit test coverage, but for now I'm just going to submit the fix without unit tests.

PR #177, opened 2019-11-07
Add an option to filter on label names

frolv:
This change adds a `source-filter` option which allows a regex to be
provided to filter the names of keys in data sources. When provided,
only keys matching the pattern will be counted as part of the reported
size. The option can be set via config files or the command line.

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -342,10 +342,16 @@ class Rollup {
     RollupRow* row = &output->toplevel_row_;
     row->vmsize = vm_total_;
     row->filesize = file_total_;
+    row->filtered_vmsize = filtered_vm_total_;
+    row->filtered_filesize = filtered_file_total_;
     row->vmpercent = 100;
     row->filepercent = 100;
     output->diff_mode_ = true;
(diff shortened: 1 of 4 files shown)


src/bloaty.cc
```diff
@@ -781,6 +817,19 @@ void RollupOutput::PrettyPrint(const OutputOptions& options,
 
   // The "TOTAL" row comes after all other rows.
   PrettyPrintRow(toplevel_row_, 0, options, out);
+
+  uint64_t filtered = 0;
+  if (ShowFile(options)) {
+    filtered += toplevel_row_.filtered_filesize;
```
haberman: I don't think adding the two together is the right approach. Your existing format is good for the case that we're showing only VM or file:

```
Filtering enabled (source_filter); omitted 15.1Mi of entries
```

For the case that both VM and file are shown lets do this instead:

```
Filtering enabled (source_filter); omitted file=12.5Mi, vm=1.2Mi of entries
```

frolv: That makes sense, updated.

PR #178, opened 2019-11-10
Recognize new WASM sections

ngg:
https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md added the DataCount section and https://github.com/WebAssembly/exception-handling/blob/master/proposals/Exceptions.md added the Event section, this PR adds support for recognizing these.

DataCount is already used by the latest Emscripten versions.


haberman:
Thank you for the contribution!

Issue #141, opened 2018-08-21
bloaty: Unexpected eh_frame format value: 8


caiiiycuk:
I tried build my binary with g++ and clang++, and analyze with bloaty. I always had same error 
```
bloaty: Unexpected eh_frame format value: 8
```

I trying to analyze elf file (linux build), when I run bloaty without flags it prints correct output:
```
     VM SIZE                         FILE SIZE
 --------------                   --------------
   0.0%       0 .debug_info        63.3Mi  73.2%
   0.0%       0 .debug_str         5.82Mi   6.7%
   0.0%       0 .debug_loc         5.06Mi   5.8%
   0.0%       0 .debug_line        3.11Mi   3.6%
  65.1%  3.09Mi .text              3.09Mi   3.6%
   0.0%       0 .debug_abbrev      1.71Mi   2.0%
   0.0%       0 .debug_ranges      1.51Mi   1.7%
   0.0%       0 .strtab             887Ki   1.0%
  17.5%   848Ki .eh_frame           848Ki   1.0%
  13.7%   664Ki .rodata             664Ki   0.7%
   0.0%       0 .symtab             340Ki   0.4%
   1.8%  86.9Ki .eh_frame_hdr      86.9Ki   0.1%
   0.0%       0 .debug_aranges     77.0Ki   0.1%
   1.0%  50.6Ki .gcc_except_table  50.6Ki   0.1%
   0.5%  25.0Ki .bss                    0   0.0%
   0.1%  6.73Ki .dynstr            6.73Ki   0.0%
   0.1%  4.50Ki [22 Others]        4.98Ki   0.0%
   0.1%  3.80Ki .dynsym            3.80Ki   0.0%
   0.1%  3.17Ki .plt               3.17Ki   0.0%
   0.0%     340 [ELF Headers]      1.86Ki   0.0%
   0.0%  1.58Ki .rel.plt           1.58Ki   0.0%
 100.0%  4.75Mi TOTAL              86.6Mi 100.0%
```

But for compileuntis or symbols I had this error.

OS: Ubuntu 16.04
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609
clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)

I tried also -g and -ggdb3 flags, same problem

haberman:
Sorry for the very slow reply. Are you still getting this error?

If you can give me a repro (like maybe post a binary somewhere) I'll debug and fix it.

[...]

PR #185, opened 2020-04-11
feat(sabsptr): support for DW_EH_PE_sabsptr

ak239:
When I run bloaty against my binary I got "bloaty: Unexpected
eh_frame format value: 8" error.
According documentation libobjc2 I can interpret this value as
signed version of DW_EH_PE_abstr.
After further research I found that this library has strange handling
for such values [2] and it mentions something about GCC extension.

It looks like this value should be returned as is, at least when
current logic tries to translate it by calling TranslateVMToFile,
this calls fails.

When this value is returned as is it looks like I am getting quite
valid output for my binary.

It might fix https://github.com/google/bloaty/issues/141

[1] https://github.com/gnustep/libobjc2/blob/master/dwarf_eh.h#L23
[2] https://github.com/gnustep/libobjc2/blob/master/dwarf_eh.h#L200

### DIFF
--- a/src/dwarf.cc
+++ b/src/dwarf.cc
@@ -1687,6 +1687,13 @@ uint64_t ReadEncodedPointer(uint8_t encoding, bool is_64bit, string_view* data,
     case DW_EH_PE_udata8:
       value = dwarf::ReadMemcpy<uint64_t>(data);
       break;
+    case DW_EH_PE_sabsptr:
+      if (is_64bit) {
+        value = dwarf::ReadMemcpy<int64_t>(data);
+      } else {
+        value = dwarf::ReadMemcpy<int32_t>(data);
+      }
+      break;
     case DW_EH_PE_sleb128:
       value = dwarf::ReadLEB128<int64_t>(data);
       break;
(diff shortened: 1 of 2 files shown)


haberman:
Thanks for the patch!

An initial comment: from what I can see, this constant is usually represented as a flag called `DW_EH_PE_signed`, not a distinct value, see:

https://github.com/gnustep/libobjc2/blob/master/dwarf_eh.h#L24
https://code.woboq.org/gcc/libgcc/unwind-pe.h.html#_M/DW_EH_PE_signed

I'll have to dig in a bit more about skipping the translation.

haberman:
Ping @ak239?

ak239:
Sorry, I missed a ping. It looks like this `DW_EH_PE_sabsptr` is a `DW_EH_PE_signed | DW_EH_PE_absptr`, so I can introduce a flag but it looks like I would need some value that presents signed absptr value (which by itself sounds to me a little bit strange).

xorz57:
Is anyone working on this?

EricRahm:
I have a rebased and slightly tweaked version of this over in #436.

EricRahm:
#436 landed, we can close this out as well.

Issue #188, opened 2020-04-26
WebAssembly DWARF support


haberman:
WebAssembly appears to support DWARF, see:

https://yurydelendik.github.io/webassembly-dwarf/#normative

https://github.com/EOSIO/eos/blob/b39b6bd3c74ff47fc53fa80aafa2c1941ba82879/CMakeModules/EosioTester.cmake.in#L25

We should see if Bloaty's existing DWARF support can be extended to support WebAssembly.

There might be some complexity here around the fact that WebAssembly doesn't have VM addresses per se.

haberman:
First mentioned in https://github.com/google/bloaty/issues/184

PR #193, opened 2020-05-07
Fix several issues when using bloaty with large numbers of input files

jchl:
Fix the following issues, which all arise when using bloaty with large numbers of input files.

1.  Signed integer overflow in a 32-bit build when the total file/VM size exceeds 2GB.
2.  An error about `--debug-file` arguments if there are more input files than threads.
3.  An assertion failure when using `--source-filter` if there are more input files than threads.
4.  Out-of-memory issues.

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -1402,7 +1402,8 @@ Bloaty::Bloaty(const InputFileFactory& factory, const Options& options)
   AddBuiltInSources(data_sources, options);
 }
 
-std::unique_ptr<ObjectFile> Bloaty::GetObjectFile(const std::string& filename) const {
+std::unique_ptr<ObjectFile> Bloaty::GetObjectFile(
+    const std::string& filename) const {
   std::unique_ptr<InputFile> file(file_factory_.OpenFile(filename));
   auto object_file = TryOpenELFFile(file);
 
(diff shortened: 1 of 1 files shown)


src/bloaty.cc
```diff
@@ -1386,57 +1389,55 @@ class Bloaty {
   std::vector<ConfiguredDataSource*> sources_;
   std::vector<std::string> source_names_;
 
-  std::vector<std::unique_ptr<ObjectFile>> input_files_;
-  std::vector<std::unique_ptr<ObjectFile>> base_files_;
-  std::map<std::string, std::unique_ptr<ObjectFile>> debug_files_;
+  std::vector<std::pair<std::string, std::string>> input_files_;
```
haberman, line 1392: Please create a struct so accesses can use descriptive names instead of `first`, `second`.

src/bloaty.cc
```diff
@@ -1386,57 +1389,55 @@ class Bloaty {
...
+                                                  bool allow_web_assembly) const {
   std::unique_ptr<InputFile> file(file_factory_.OpenFile(filename));
   auto object_file = TryOpenELFFile(file);
 
   if (!object_file.get()) {
     object_file = TryOpenMachOFile(file);
   }
 
-  if (!object_file.get()) {
+  if (!object_file.get() && allow_web_assembly) {
```
haberman, line 1411: Is there a need to specifically disallow debug files with WebAssembly here? There is a feature request to support DWARF with WebAssembly: https://github.com/google/bloaty/issues/188

src/bloaty.cc
jchl: Will do.

src/bloaty.cc
jchl: I was wondering that myself, but I honestly have no idea what a WebAssembly file is, so I was just preserving the previous behavior.

If there is no need for this, then I'm very happy to remove that extra complexity.

haberman:
Thanks for these fixes, this looks great now.

Do you think any of these fixes have reasonably easy ways to verify in unit tests?

PR #207, opened 2020-06-28
Updated ABSL and finally used ABSL's CMake build system.

haberman:


EricRahm:
We did this over in #347.

PR #211, opened 2020-08-17
[dwarf] Handle tombstone address values

tengyifei:
These were introduced recently in LLVM, and indicates discarded code (see https://reviews.llvm.org/D81784).

Manually tested that bloaty runs successfully on binaries containing tombstone values.

cc @MaskRay

### DIFF
--- a/src/dwarf.cc
+++ b/src/dwarf.cc
@@ -18,6 +18,7 @@
 #include <algorithm>
 #include <initializer_list>
 #include <iostream>
+#include <limits>
 #include <memory>
 #include <stack>
 #include <unordered_map>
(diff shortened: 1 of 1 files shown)


src/dwarf.cc
```diff
@@ -60,6 +61,15 @@ int DivRoundUp(int n, int d) {
   return (n + (d - 1)) / d;
 }
 
+namespace {
+
+// uint64/32 max is a tombstone value added by https://reviews.llvm.org/D81784.
+bool IsValidDwarfAddress(uint64_t addr, bool dwarf64) {
+  return addr != 0 && addr != (dwarf64 ? std::numeric_limits<uint64_t>::max()
```
MaskRay, line 68: DWARF64 actually refers to a different encoding format. It is orthogonal to the address space size of the host.

You'll want to check `address_size == 4`

tengyifei: Done, PTAL

haberman:
Thank you for the PR and review!

PR #218, opened 2020-11-13
Add option to not rely on RE2

tweksteen:
When disabling RE2, using the parameter --source-filter or custom_data_source will raise an error.


haberman:
CMake changes look good. I'd prefer if we could minimize the number of `#ifdefs` in the source code. Could we instead define a little wrapper around RE2, eg.

```
#ifdef USE_RE2
class Regex {
  Regex(pattern) : re2_(pattern) {}
  Extract(...) { RE2::Extract(re2_, ...);
  // etc.
 private:
  RE2 re2_;
}
#else
class Regex {
  Regex(pattern) : re2_(pattern) { abort(); }
  Extract(...) {}
}
#endif
```

Then we can add just a few #ifdefs in the code where we need to warn the user that the feature is disabled. Does that make sense?

tweksteen:
Thanks. Done. I created a new `re.h` file which contains the 2 implementations. I used `ReImpl` for the new class name to avoid conflict with the proto class `Regex`.

PR #219, opened 2020-12-03
Allow building with MSVC 2019

learn-more:


src/bloaty.cc
```diff
@@ -914,24 +920,61 @@ class MmapInputFile : public InputFile {
   BLOATY_DISALLOW_COPY_AND_ASSIGN(MmapInputFile);
 };
 
+#if defined(_MSC_VER)
```
haberman, line 923: Instead of `#ifdef`ing inside `FileDescriptor` and `MmapInputFile`, I'd prefer to use separate classes. `InputFile` is a virtual base, so we can create a different derived class `WindowsInputFile`.

Generally I prefer to keep `#ifdef` to a minimum.

haberman (changes requested):
Sorry for the slow review, I missed this when it first came in.

learn-more:
> Sorry for the slow review, I missed this when it first came in.

No problem!

I do have another branch where I setup appveyor to build the msvc version, shall I include that in the update as well?

haberman:
> I do have another branch where I setup appveyor to build the msvc version, shall I include that in the update as well?

That sounds like a good idea. Does this require any changes to the repo config?

learn-more:
> > I do have another branch where I setup appveyor to build the msvc version, shall I include that in the update as well?
> 
> That sounds like a good idea. Does this require any changes to the repo config?

There is a script (that I'll add):
https://github.com/learn-more/bloaty/commit/20acde975e3118497e90b39d4c944e3ad56e9079

Then you just have to login on appveyor with your account, and activate this repository (can be done per-repo or account wide).
After that, you can add a badge to the readme, or collect files as artifacts (if you want to add binary releases for windows users, for example)

src/bloaty.cc
learn-more: In that case I would have to ifdef out the `MmapInputFile` and `FileDescriptor` entirely on windows (I'll use `_MSC_VER`, unless you have a preference for `_WIN32` instead?)

haberman: Yes, I think that's a good approach. Later we could split off the UNIX or Windows-specific code into separate files like `unix.cc` and `windows.cc`, especially if the amount of platform-specific code grows, but I don't think that's necessary now.

learn-more:
The suggestions have been incorporated, please let me know if there is anything else that you'd like me to change.

ps: It might be a good idea to include a `.clang-format` file fixing the exact clang-format configuration.

learn-more:
@haberman Would you have time to take a look at this again?

haberman:
Sorry for the delay! I am not exactly clear how to get AppVeyor going, but I'll merge this to unblock you.

learn-more:
> Sorry for the delay! I am not exactly clear how to get AppVeyor going, but I'll merge this to unblock you.

Thanks!

Basically, you have to login on [appveyor](https://www.appveyor.com/) (using your github account is the easiest).
Then you can enable it for this repository by clicking the 'New project' button,
and either giving it access to all your repositories, or only this one.
https://www.appveyor.com/docs/

PR #230, opened 2021-04-21
Support zlib-gnu compressed debug symbols

### LINKED ISSUE #204: Go compatibility: implement compressed debug info
Go, since version 1.11 (released 2018), compresses debug information by default.

This makes some bloaty features not work by default, e.g. breaking down size by compilation unit.

As a workaround, you can disable debug information compression in Go by using the `-ldflags=-compressdwarf=false` flag when running `go install`.

Ideally, though, bloaty would implement support for reading compressed debug info and work with Go binaries out of the box :)

related: https://github.com/google/bloaty/issues/42

aarzilli:
Supports debug symbols compressed in the zlib-gnu standard (i.e.
zdebug).

Fixes #204

--

Please note the memory leak.


haberman:
Thanks for the PR. There are some things I'd like to clean up (like the memory leak), but I can do that after the PR is submitted.

Issue #203, opened 2020-05-26
inverted --source-filter


grandinj:
awesome tool!

could we get an inverted --source-filter please? i.e. to exclude stuff from the report.

because it appears to be unreasonably hard to generate a negative regular expression that RE2 will accept.

For example, trying to exclude c++ std library classes:

    bloaty -d shortsymbols,fullsymbols --source-filter ^[^s][^t][^d] instdir/program/libsclo.so

haberman:
Yes I think this is a great idea.

Biggest question is how to specify this on the command line. The first things that come to mind are:

```
bloaty --source-filter !std
```

or:

```
bloaty --source-filter ~std
```

But unfortunately both of those have special meanings for the shell.

grandinj:
git has --not e.g. see

   https://git-scm.com/docs/git-grep#Documentation/git-grep.txt---and

--source-filter-not
or
--not-source-filter

also works for me.

Or convince the RE people to support the perl
  ?!
operator :-)

vinuraja:
+1. I was recently working with `bloaty` and this is a feature I'm really missing.

PR #234, opened 2021-05-03
Add an option to exclude based on source names

learn-more:
#203

This includes a testcase for both include and exclude.

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -338,8 +338,9 @@ class Rollup {
     CreateRows(row, base, options, true);
   }
 
-  void SetFilterRegex(const ReImpl* regex) {
-    filter_regex_ = regex;
+  void SetFilterRegex(const ReImpl* regex_include, const ReImpl* regex_exclude) {
+    filter_regex_include_ = regex_include;
+    filter_regex_exclude_ = regex_exclude;
   }
 
   // Subtract the values in "other" from this.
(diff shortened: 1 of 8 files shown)


learn-more:
@haberman apart from the merge conflict, anything here that should be improved?

src/bloaty.cc
```diff
@@ -1940,6 +1959,8 @@ USAGE: bloaty [OPTION]... FILE... [-- BASE_FILE...]
   --list-sources     Show a list of available sources and exit.
   --source-filter=PATTERN
                      Only show keys with names matching this pattern.
+  --not-source-filter=PATTERN
+                     Exclude keys with names matching this pattern.
```
haberman, line 1963: We should document what happens if a name matches both `--source-filter` and `--exclude-source-filter`.

src/bloaty.cc
```diff
@@ -1940,6 +1959,8 @@ USAGE: bloaty [OPTION]... FILE... [-- BASE_FILE...]
   --list-sources     Show a list of available sources and exit.
   --source-filter=PATTERN
                      Only show keys with names matching this pattern.
+  --not-source-filter=PATTERN
```
haberman, line 1962: I'm leaning towards `--exclude-source-filter`. It matches the rest of the PR's terminology better.

haberman (changes requested):
This looks great, just a few comments.

learn-more:
@haberman I'd like to finish this after #264 is merged, so that I can implement this testcase with yaml2obj instead of the current c++ test. (I will remove the c++ test from this PR, and add a lit test)

src/bloaty.cc
learn-more: I wrote a test that shows what happens, basically nothing is output other than the fact that everything is omitted.

learn-more: This has been added the the help text.

learn-more:
@haberman sorry for the delay, I finally updated this.
The tests fail due to #317

PR #244, opened 2021-07-30
PE: use an `enum class` rather than `namespace` (NFC)

compnerd:
This moves the PE size constants into an enumeration class, which is a
C++ 11 construct, rather than a namespace.  This enforces the explicit
references and adds stronger type checking.

### DIFF
--- a/src/pe.cc
+++ b/src/pe.cc
@@ -23,39 +23,43 @@ namespace pe {
 const uint16_t dos_magic = 0x5A4D;  // MZ
 
 //! Sizes in bytes of various things in the COFF format.
-namespace STRUCT_SIZES {
-enum {
-  Header16Size = 20,
-  Header32Size = 56,
-  NameSize = 8,
-  Symbol16Size = 18,
-  Symbol32Size = 20,
-  SectionSize = 40,
-  RelocationSize = 10,
-  BaseRelocationBlockSize = 8,
-  ImportDirectoryTableEntrySize = 20,
-  ResourceDirectoryTableSize = 16,
-  ResourceDirectoryEntriesSize = 8,
-  ResourceDataEntrySize = 16
+enum class StructSizes {
+  kHeader16Size = 20,
+  kHeader32Size = 56,
+  kNameSize = 8,
+  kSymbol16Size = 18,
+  kSymbol32Size = 20,
+  kSectionSize = 40,
+  kRelocationSize = 10,
+  kBaseRelocationBlockSize = 8,
+  kImportDirectoryTableEntrySize = 20,
+  kResourceDirectoryTableSize = 16,
+  kResourceDirectoryEntriesSize = 8,
+  kResourceDataEntrySize = 16,
 };
-}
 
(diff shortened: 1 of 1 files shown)


learn-more:
Now all places where these items are used need casts....
What does this fix?

compnerd:
@learn-more ah, this was to repair the internal builds.  However, the use of namespace + struct is better modeled as an enum class.  Id say that this is most likely best handled via `constexpr` values.  I would be fine with changing that now that everything actually builds again.

learn-more:
> @learn-more ah, this was to repair the internal builds. However, the use of namespace + struct is better modeled as an enum class. Id say that this is most likely best handled via `constexpr` values. I would be fine with changing that now that everything actually builds again.

'repair the internal builds' ?
Is there some code-checker that prohibits namespace with an anonymous enum?

Because clearly strong typing here is no argument, considering the initial enum was anonymous and thus (by design) could not be used as variable.

Sorry, but I fail to see any gains from this change.

haberman:
This change came from me, at the suggestion of another internal reviewer, when I imported this code into Google a couple months ago.

I can see how the anonymous-enum-inside-namespace had a certain elegance to it, in that it allows names like `STRUCT_SIZES::SectionSize` without casts and without allowing the enum to be instantiated.

That said, this is a somewhat unusual pattern (at least in Google code) so readers won't know the rationale. We try to follow common idioms in Google code to help readers, since code is read much more often than it is written.

We could change this to something like:

```
constexpr size_t kSectionSize = 40;
static_assert(sizeof(pe_section) == kSectionSize, "Compiler options broke LIEF struct layout");
```

Or even just:

```
// Assumed to be section size, since we're comparing against sizeof(pe_section)
static_assert(sizeof(pe_section) == 40, "Compiler options broke LIEF struct layout");
```

learn-more:
> This change came from me, at the suggestion of another internal reviewer, when I imported this code into Google a couple months ago.
> 
> I can see how the anonymous-enum-inside-namespace had a certain elegance to it, in that it allows names like `STRUCT_SIZES::SectionSize` without casts and without allowing the enum to be instantiated.
> 
> That said, this is a somewhat unusual pattern (at least in Google code) so readers won't know the rationale. We try to follow common idioms in Google code to help readers, since code is read much more often than it is written.

I would argue that the static_cast is also not an improvement for reading, but at least I understand the idea behind the change now.
(I am less concerned about writing the static cast, more about reading it btw, I fully agree with the only written once, read alot)

> 
> We could change this to something like:
> 
> ```
> constexpr size_t kSectionSize = 40;
> static_assert(sizeof(pe_section) == kSectionSize, "Compiler options broke LIEF struct layout");
> ```
> 
> Or even just:
> 
> ```
> // Assumed to be section size, since we're comparing against sizeof(pe_section)
> static_assert(sizeof(pe_section) == 40, "Compiler options broke LIEF struct layout");
> ```

The original was taken 1:1 from `LIEF/include/LIEF/PE/Structures.hpp.in`, but from those 2 cases I would prefer the constexpr.

PR #249, opened 2021-07-30
Update build.yml

compnerd:
The `workflow_dispatch` event is used for triggering manual jobs, `workflow_jobs` is for consumption of other jobs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,20 +4,21 @@ on:
   pull_request:
     branches:
       - master
-  workflow_run:
+  workflow_dispatch:
 
 jobs:
   windows:
     runs-on: windows-latest
 
     strategy:
+      fail-fast: false
       matrix:
-        arch: ["Win32", "x64", "ARM64"]
+        arch: ["Win32", "x64"] # "ARM64"
 
     steps:
-      - uses: actions/echeckout@v2
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


haberman:
Looks like the Windows build failed: https://github.com/google/bloaty/pull/249/checks?check_run_id=3206306907

compnerd:
Yeah, the problem is that the build configuration currently is doing some atrocious things to deal with static linking on Windows (which goes against the recommendation of upstream CMake).  There is a way to fix this, but the change requires that it propagates into protobuf, which makes it a bit more challenging.

compnerd:
Restored the 32-bit build on Windows as #252 should repair the 32-bit builds.

haberman:
> There is a way to fix this, but the change requires that it propagates into protobuf, which makes it a bit more challenging.

I'm on the protobuf team. If there's something in our CMake that should be improved, I can help land it.

compnerd:
> If there's something in our CMake that should be improved, I can help land it.

Thanks for the offer ...  and since no good deed goes unpunished, expect an incoming PR for that :)

compnerd:
ooh, and I think I figured out how to make it work without the needed change (but we should still follow up on the protobuf change just to improve protobufs).

compnerd:
Okay, the testing isn't working yet, the builds are.  I guess that's a step forward.

Issue #221, opened 2020-12-17
Expand/improve Bloaty's unit tests


haberman:
Bloaty's test coverage is not great. Bloaty's existing tests use the output of a C compiler as their input, and then try to assert things about what the sizes should be. But this model is inherently fragile, because C compilers have a lot of latitude in what kind of object files and binaries they produce.

A better way to test Bloaty would be to use assembly as the input. This can give us much more control over the object file, which can help our unit tests be far more precise.

Bloaty is target-independent: a single Bloaty binary can analyze ELF or Mach-O. Ideally our tests can be cross-platform also, so you can test all object file formats from a single platform. Currently I am leaning towards using [NASM](https://www.nasm.us/) as the assembler, because it is also target-independent:

```
# Produce a Mach-O object file
$ nasm -fmacho64 test.asm
# Produce an ELF object file
$ nasm -felf64 test.asm
```

This is very attractive, and nasm is also small and self-contained. On my machine it compiles in less than 4 seconds of real time:

```
$ time make -j72
real    0m3.366s
user    0m23.991s
sys     0m8.680s
```

Linking is trickier. I'm not aware of any target-independent linkers that can link both ELF and Mach-O from a single linker binary. I did find [this portable port of the Apple linker](https://github.com/tpoechtrager/cctools-port) which seems to be quite portable. The "gold" linker is ELF native, and should be portable, but it's tied up in the bigger and more complicated "binutils" package. Not sure the best solutions here.

haberman:
After some more research, I'm not sure if NASM will cut it. There are some important features it doesn't appear to support.

- Unwind Information: `.eh_frame` contains unwind information, and it appears this is generated somewhat automatically by the [assembler from directives named `.cfi_*`](https://sourceware.org/binutils/docs/as/CFI-directives.html#CFI-directives), eg. `. cfi_startproc `, `.cfi_endproc`, etc. NASM appears to contain no support for this.
- DWARF `.loc` directives, for emitting `.debug_line`.

Besides this, NASM format is less convenient to use because you cannot get the compiler to generate it manually (the compiler's output is meant for `gas`) and NASM format isn't as common or well known.

Perhaps it is better to use the toolchain assembler and linker, but check in the output files so that all formats can be tested no matter what platform one is developing on.

The main open question then is whether the assembler's output is deterministic enough for this to work.

haberman:
[...]

PR #250, opened 2021-07-31
DWARF: address a truncation warning

compnerd:
The result of a `size` was being stored into a `std::pair` where the
type was being stored as a `uint32_t`.  However, the result of a `size`
to a STL container is a `size_t`.  This extends the size member to the
proper size, avoiding a warning.

### DIFF
--- a/src/dwarf.cc
+++ b/src/dwarf.cc
@@ -607,7 +607,7 @@ class DIEReader {
   // both the current abbrev. table and the sizes.
   uint32_t abbrev_version_;
 
-  std::map<std::pair<AbbrevTable*, CompilationUnitSizes>, uint32_t>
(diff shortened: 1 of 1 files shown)


haberman:
Thanks for this. Just as a heads-up, I have a large refactoring of the DWARF code in progress:

https://github.com/google/bloaty/compare/master...haberman:refactor?expand=1

I want to land this refactoring, but it has some regressions I haven't gotten to the bottom of yet.

That led me to focus my energies on addressing https://github.com/google/bloaty/issues/221, which I've made some progress on https://github.com/google/bloaty/compare/master...haberman:better-tests?expand=1

compnerd:
Noted, unfortunately, I do have one more similarly inspired change for the tests.  I couldn't process the actual failure on the Windows builds because the truncation is so deeply nested it causes pages of warnings.

PR #253, opened 2021-07-31
build: repair the build on Windows

compnerd:
CMake 3.15 introduced a proper top-level option to control the MSVC
runtime programatically on a per-target basis.  The
`MSVC_RUNTIME_LIBRARY` target property permits control.over the runtime
library on a single target.  This value is automatically initialized to
the value of `CMAKE_MSVC_RUNTIME_LIBRARY` if the variable is set.
Prefer to use this mechanism for controlling the library rather than
processing the *user* controlled compile flags.  This corrects the
library mismatch between protoc and the bloaty binaries when building on
Windows.


compnerd:
The diff here looks worse than it is.  The majority of it is whitespace changes due to indentation - viewing ignoring whitespace changes makes it easier to see what is being changed.

PR #254, opened 2021-08-01
CI: adjust the testing command

compnerd:
This adjusts the testing command to match the existing Travis CI
invocation of "make test".  Use `cmake` to drive the test target
to allow using a different make program more easily.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: build
         run: cmake --build build --config Debug
       - name: test
-        run: ctest build
+        run: cmake --build build --config Debug --target RUN_TESTS
 
   macOS:
     runs-on: macos-latest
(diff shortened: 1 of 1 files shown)


compnerd:
The multithreaded bloaty test is pretty heavy on Windows, it seems like its just hanging when its just really just slow.

haberman:
Hmm, I wonder why it's so much slower on Windows. That's a little concerning.

Perhaps we should decrease this parameter to let the test pass on Windows: https://github.com/google/bloaty/blob/8200c2972dd163a3cb8d56ef182a375a43e14e1e/tests/bloaty_misc_test.cc#L47

compnerd:
Yeah, I do wonder why its so much slower.  It does pass, it jus takes a *long* while (as you can see from the times - ~30m).

haberman:
Oh, it looks like the ultimate failure was a timeout in fuzz_test:

```
7/8 Test #7: fuzz_test ........................***Timeout 1500.01 sec
```

Was fuzz_test fixed by https://github.com/google/bloaty/pull/255?

compnerd:
Yeah, #255 repaired that timeout.  I'm going to rebase this to retest.

haberman:
Windows build is passing in 8-9 minutes now, that's fantastic. I'll merge this now.

PR #259, opened 2021-08-02
tests: convert the test suite to use more of the LLVM test utilities

compnerd:
This converts the project from the custom test runner to use the LLVM
test tools.  While this is still not ready to integrate into the test
runs on CI, this is slowly working towards that polish.


haberman:
@googlebot I consent

PR #260, opened 2021-08-03
CI: build test dependencies in CI for auditable builds

compnerd:
The dependencies should be relatively light and allow us to have an
audit trail for the software.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,12 +17,40 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+        with:
+          repository: llvm/llvm-project
+          ref: llvmorg-13.0.0-rc1
+          path: llvm-project
+          fetch-depth: 1
+      - name: Get LLVM Revision
+        id: llvm-revision
+        run: |
+          echo "::set-output name=revision::$(git -C ${{ github.workspace }}/llvm-project rev-parse HEAD)"
+      - uses: actions/cache@v2
+        id: llvm-build
+        with:
+          path: ${{ github.workspace }}/build/llvm-project
+          key: ${{ runner.os }}-${{ matrix.arch }}-${{ steps.llvm-revision.outputs.revision }}
+      - name: configure LLVM
+        if: steps.llvm-build.outputs.cache-hit != 'true'
+        run: cmake -B ${{ github.workspace }}/build/llvm-project -D CMAKE_BUILD_TYPE=Release -G "Visual Studio 16 2019" -S ${{ github.workspace }}/llvm-project/llvm
+      - name: build LLVM tools
+        if: steps.llvm-build.outputs.cache-hit != 'true'
+        run: |
+          cmake --build ${{ github.workspace }}/build/llvm-project --target FileCheck
+          cmake --build ${{ github.workspace }}/build/llvm-project --target yaml2obj
+
+      - uses: actions/checkout@v2
+        with:
+          path: bloaty
       - name: configure
-        run: cmake -B build -D CMAKE_BUILD_TYPE=Debug -G "Visual Studio 16 2019" -A ${{ matrix.arch }} -S .
+        run: cmake -B build/bloaty -D CMAKE_BUILD_TYPE=Debug -G "Visual Studio 16 2019" -A ${{ matrix.arch }} -S ${{ github.workspace }}/bloaty -D FILECHECK_EXECUTABLE=${{ github.workspace }}/build/llvm-project/bin/FileCheck.exe -D YAML2OBJ_EXECUTABLE=${{ github.workspace }}/build/llvm-project/bin/yaml2obj.exe -D LIT_EXECUTABLE=${{ github.workspace }}/llvm-project/utils/lit/lit.py
       - name: build
-        run: cmake --build build --config Debug
+        run: cmake --build build/bloaty --config Debug
       - name: test
-        run: cmake --build build --config Debug --target RUN_TESTS
+        run: |
+          cmake --build build/bloaty --config Debug --target RUN_TESTS
+          cmake --build build/bloaty --config Debug --target check-bloaty
 
   macOS:
     runs-on: macos-latest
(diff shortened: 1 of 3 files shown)


haberman:
This is great!

This appears to be Windows-only for now, is that intentional?

Also I can't tell if the `lit` tests ran or not from the logs. I can definitely see that `yaml2obj.exe` was built and cached.

haberman:
Also it looks like Win32 and Win64 conflicted with each other for the cache keys:

```
Unable to reserve cache with key Windows-, another job may be creating this cache.
```

compnerd:
Yes, the Windows-only build was intentional.  I wanted to get a rough idea of the pattern to use to get the things wired together.  This is currently purely building the dependencies.  Once all of the platforms have the tools available, I can tweak the build to start running the tests.

.github/workflows/build.yml
```diff
@@ -17,12 +17,40 @@ jobs:
...
+          path: ${{ github.workspace }}/build/llvm-project
+          key: ${{ runner.os }}-${{ matrix.arch }}-${{ steps.llvm-revision.outputs.revision }}
+      - name: configure LLVM
+        if: steps.llvm-build.outputs.cache-hit != 'true'
+        run: cmake -B ${{ github.workspace }}/build/llvm-project -D CMAKE_BUILD_TYPE=Release -G "Visual Studio 16 2019" -S ${{ github.workspace }}/llvm-project/llvm
+      - name: build LLVM tools
+        if: steps.llvm-build.outputs.cache-hit != 'true'
+        run: |
+          cmake --build ${{ github.workspace }}/build/llvm-project --config Release --target FileCheck
+          cmake --build ${{ github.workspace }}/build/llvm-project --config Release --target yaml2obj
```
learn-more, line 41: Would it be an idea to also create artifacts from this, so contributors can download those and configure it locally?
When going that route, I would suggest to also include `obj2yaml` so that new tests can be added somewhat simpler.

learn-more: Or even to create a new build definition,
only triggered manually that builds the llvm binaries, and attached it to a 'Tools' release in this repo.
Then this tools release can be checked out in the builder, so that it does not have to rely on the github cache, and humans can easily grab the used tooling as well.
This will also ensure that the used tooling does not change overnight when there is a cache miss.

CMakeLists.txt
```diff
@@ -335,15 +335,20 @@ else()
...
+    if(Python_FOUND AND LIT_EXECUTABLE AND FILECHECK_EXECUTABLE AND YAML2OBJ_EXECUTABLE)
       set(BLOATY_SRC_DIR ${PROJECT_SOURCE_DIR})
       set(BLOATY_OBJ_DIR ${PROJECT_BINARY_DIR})
       configure_file(tests/lit.site.cfg.in tests/lit.site.cfg @ONLY)
 
-      # TODO(abdulras) polish this further and convert this to `add_test` so that
-      # the default test suite run executes the tests.
       add_custom_target(check-bloaty
-        COMMAND ${Python_EXECUTABLE} ${LIT_EXECUTABLE} -sv ${PROJECT_BINARY_DIR}/tests
+        COMMAND ${Python_EXECUTABLE} ${LIT_EXECUTABLE} -sv ${PROJECT_BINARY_DIR}/tests --param bloaty=$<TARGET_FILE:bloaty>
```
learn-more, line 347: The param is used for the `Debug` or `Release` part from the build indeed, but `LIT_EXECUTABLE` will be `lit.exe` on windows, which means you cannot pass it to python.

CMakeLists.txt
```diff
@@ -335,15 +335,20 @@ else()
     enable_testing()
 
     find_package(Python COMPONENTS Interpreter)
-    if(Python_FOUND)
+    find_program(LIT_EXECUTABLE NAMES lit lit.py)
```
learn-more, line 338:
```suggestion
    find_program(LIT_EXECUTABLE NAMES lit-script.py lit.py lit)
```
If you change it to this (and do `pip install lit` in the build definition) it _might_ work on windows

CMakeLists.txt
compnerd: `lit` is definitely a python script, the tests are already executing it here.

CMakeLists.txt
compnerd: The tests are working on Windows already :). Adding another spelling is fine.

haberman: It appears this PR is already successfully running the `lit` tests on Windows: https://github.com/google/bloaty/pull/260/checks?check_run_id=3243382953

learn-more: hm, I could have sworn that they were red, guess I looked at old or wrong builds

.github/workflows/build.yml
learn-more: (and bring build times down by ~10 minutes for cache misses)

CMakeLists.txt
learn-more: On my dev machine this does not work:
```
5>Running bloaty tests...
5>SyntaxError: Non-UTF-8 code starting with '\x90' in file C:\Program Files\Python39\Scripts\lit.exe on line 1, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
5>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(238,5): error MSB8066: Custom build for 'E:\Projects\bloaty\build\CMakeFiles\16d0e534f5107e26a948780ce7ba2662\check-bloaty.rule' exited with code 1.
5>Done building project "check-bloaty.vcxproj" -- FAILED.
```

CMakeLists.txt
```diff
@@ -335,15 +335,20 @@ else()
...
       add_custom_target(check-bloaty
-        COMMAND ${Python_EXECUTABLE} ${LIT_EXECUTABLE} -sv ${PROJECT_BINARY_DIR}/tests
+        COMMAND ${Python_EXECUTABLE} ${LIT_EXECUTABLE} -sv ${PROJECT_BINARY_DIR}/tests --param bloaty=$<TARGET_FILE:bloaty>
+        DEPENDS
+          bloaty
+          ${CMAKE_CURRENT_SOURCE_DIR}/tests/lit.cfg
+          ${CMAKE_CURRENT_BINARY_DIR}/tests/lit.site.cfg
         COMMENT "Running bloaty tests..."
         USES_TERMINAL)
     endif()
```
learn-more, line 354:
```suggestion
      set_property(TARGET check-bloaty PROPERTY FOLDER "tests")
    endif()
```
This groups it nicely with the other tests in the IDE

CMakeLists.txt
learn-more: For my dev machine to pick this up, `lit-script.py` needs to come before the `lit` entry.

PR #261, opened 2021-08-04
Update lit config / usage to be msvc compatible

learn-more:
* `lit`, `filecheck` and `yaml2obj` will be auto-detected using find_program now. They can still be configured manually
* Added the `BLOATY_` prefix on the variables to configure the paths
* Added the lit config files to the `SOURCES` of `check-bloaty` so that they show up in the IDE
* Passed the bloaty path as parameter to `lit`, so that it works with Visual Studio's multi-config

@compnerd Can you please verify that it still works for you?
Note, I have updated the variables to include the `BLOATY_` prefix.

--- 

MSVC output when building `check-bloaty`:
```
Build started...
1>------ Build started: Project: check-bloaty, Configuration: Debug x64 ------
1>Running bloaty tests...
1>-- Testing: 6 tests, 6 workers --
1>Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90..
1>
1>Testing Time: 0.43s
1>  Passed: 6
========== Build: 1 succeeded, 0 failed, 9 up-to-date, 0 skipped ==========
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,14 +338,24 @@ else()
     if(Python_FOUND)
       set(BLOATY_SRC_DIR ${PROJECT_SOURCE_DIR})
       set(BLOATY_OBJ_DIR ${PROJECT_BINARY_DIR})
+
+      find_program(BLOATY_LIT_EXECUTABLE lit lit.exe DOC "Specifies where to find the lit tool" REQUIRED)
+      find_program(BLOATY_FILECHECK_EXECUTABLE filecheck FileCheck.exe DOC "Specifies where to find the lit tool" REQUIRED)
+      find_program(BLOATY_YAML2OBJ_EXECUTABLE yaml2obj yaml2obj.exe DOC "Specifies where to find the yaml2obj tool" REQUIRED)
+
       configure_file(tests/lit.site.cfg.in tests/lit.site.cfg @ONLY)
 
       # TODO(abdulras) polish this further and convert this to `add_test` so that
       # the default test suite run executes the tests.
       add_custom_target(check-bloaty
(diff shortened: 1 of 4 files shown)


learn-more:
Ok, everything fails because I have marked the binaries as `REQUIRED`, so maybe it was not quite ready for that :)

compnerd:
@learn-more thanks for the change.  This was already covered in #260.  I don't know about the `BLOATY_` prefix on the tool names, as that tends to be how most of the tools/packages are specified (without the prefix).  In a discussion with @haberman, we decided to keep the extended tests optional for now, but wanted to get that enabled on CI.  Do the changes in #260 solve your issues?

learn-more:
> @learn-more thanks for the change. This was already covered in #260. I don't know about the `BLOATY_` prefix on the tool names, as that tends to be how most of the tools/packages are specified (without the prefix). In a discussion with @haberman, we decided to keep the extended tests optional for now, but wanted to get that enabled on CI. Do the changes in #260 solve your issues?

The `BLOATY_` prefix enables grouping in the cmake GUI:
image

What I also noticed in the testing is that lit on windows will be `lit.exe`, which cannot be run like `python lit.exe other arguments`, since python tries to run this as `.py` file

compnerd:
I'm not sure I follow the `lit` bit - `lit` is a python script, not an executable.  How are you finding that it is an executable?

learn-more:
> I'm not sure I follow the `lit` bit - `lit` is a python script, not an executable. How are you finding that it is an executable?

On windows, when doing `pip install lit` it will install a `lit.exe` that runs a python script.
Next to this file is `lit-script.py` so we can probably search for that first.

Issue #220, opened 2020-12-05
Direction for PE file support


learn-more:
To add support for PE files there are a few different approaches that can be used:

- Import a submodule for [LIEF](https://github.com/lief-project/LIEF) (this also supports other files, maybe all parsers could leverage this?)
- Import a submodule for a parser like [pe-parse](https://github.com/trailofbits/pe-parse) which is more of an iterating parser, and specialized for PE files
- Grab some headers / classes from LLVM and write something along the lines of [COFFDump](https://github.com/llvm/llvm-project/blob/master/llvm/tools/llvm-objdump/COFFDump.cpp)
- Grab only the [PE headers](https://github.com/llvm/llvm-project/blob/master/llvm/include/llvm/BinaryFormat/COFF.h) from LLVM and write a complete custom parser 
- Other (please specify)

What would be the preferred way of moving forward?

haberman:
Thanks for your interest in adding PE support! This is something I've wished for for a while.

Generally with Bloaty I have found that custom parsers are necessary. Bloaty cares about not only the data in the file, but the *precise location* of each bit of data in the file. For example, for the file headers and symbol table entries, we need to not only read them, but report their byte range within the file.

Generally I've found that existing libraries do not offer this information, because almost no program besides Bloaty needs it. For this reason, all of the existing parsers in Bloaty take the final approach you mentioned (grab the headers and write a complete custom parser). I expect PE will probably require the same.

learn-more:
@haberman now that the initial PR is merged,
how do you want to proceed with PE support?

haberman:
Now that we have the `lit` testing in place, I'm a lot more comfortable moving forward with expanding PE support.

I'd love to see support for:

- `segments`: this would be the regions of the file that the loader will load. The `segments` name is somewhat ELF-specific, but I think PE has something similar, like in the [optional header](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-image-only)?
- `symbols`: using the symbol table hopefully we could get some good symbol support here.
- `compileunits`: I assume PE files have this information available for debugging?

What do you think?

learn-more:
> Now that we have the `lit` testing in place, I'm a lot more comfortable moving forward with expanding PE support.
> 
> I'd love to see support for:
> 
[...]

PR #264, opened 2021-08-04
Add some simple PE lit testcases

learn-more:
All testcases compiled with tcc/tcc64
the 'empty' PE files do as less as possible (so no imports, no exports, close to no code, just return 1)
the 'simple' PE files call puts, and the dll's have an additional export (and a bunch of internal symbols exported)


tests/PE/sections/empty-pe32-dll.test
```diff
@@ -0,0 +1,88 @@
...
+  - Name:            .text
+    Characteristics: [ IMAGE_SCN_CNT_CODE, IMAGE_SCN_MEM_EXECUTE, IMAGE_SCN_MEM_READ ]
+    VirtualAddress:  4096
+    VirtualSize:     24
+    SectionData:     5589E581EC0000000090B801000000E900000000C9C20C00
+symbols:         []
+...
+
+# CHECK: FILE MAP:
+# CHECK: 000-1a0	        416		[PE Headers]
```
haberman, line 80: For ELF we try to attribute section headers to their corresponding sections. For example, the section header for `.text` is attributed to `.text` instead of `[ELF Headers]`.

Would this be possible for PE?

haberman (commented):
These are great, thanks!

tests/PE/sections/empty-pe32-dll.test
learn-more: As outlined in #220, it can be split into:
* DOS Header
* DOS Stub
* Rich Header? (not sure if this can be done reliable, otherwise it will be dumped on top of the DOS Stub)
* NT header
* Per-section header (for each .text, .rdata etc section)

Do you just want the Per-section header here, or also the other DOS/NT header split?

learn-more: and do you want that only for `segments`, or also for `sections` ?

haberman: I think mainly the per-section headers should be attributed to their respective sections. The other headers sound like they are not specific to any particular section, so `[PE Headers]` sounds like a good fallback for them.

For ELF I've recently changed segments to work the same way: each segment header is attributed to its respective segment.

learn-more: Option 1:
```
Maps for empty-pe64-exe.test.tmp.obj:

FILE MAP:
000-188         392             [PE Headers]
188-1b0          40             .text Header
1b0-1d8          40             .pdata Header
1d8-200          40             [Unmapped]
200-400         512             .text
400-600         512             .pdata

VM MAP:
0000-0188               392             [PE Headers]
0188-01b0                40             .text Header
01b0-01d8                40             .pdata Header
01d8-1000              3624             [-- Nothing mapped --]
1000-1020                32             .text
1020-2000              4064             [-- Nothing mapped --]
2000-200c                12             .pdata
```

Or option 2:
```
Maps for empty-pe64-exe.test.tmp.obj:

FILE MAP:
000-040          64             DOS Header
040-080          64             DOS Stub
080-188         264             NT Header
188-1b0          40             .text Header
1b0-1d8          40             .pdata Header
1d8-200          40             [Unmapped]
200-400         512             .text
400-600         512             .pdata

VM MAP:
0000-0040                64             DOS Header
0040-0080                64             DOS Stub
0080-0188               264             NT Header
0188-01b0                40             .text Header
01b0-01d8                40             .pdata Header
01d8-1000              3624             [-- Nothing mapped --]
1000-1020                32             .text
1020-2000              4064             [-- Nothing mapped --]
2000-200c                12             .pdata
```

haberman: Thanks for trying these out. After playing with the ELF support, I've change my mind on this (sorry for the change). I think I want to change ELF to use:

- [ELF Header] (for the fixed-size ELF header at the top of the file)
- [ELF Section Headers] (used for all section headers)
- [ELF Program Headers] (used for all program headers)

For PE I think this would mean:

- [PE Headers] (for DOS and NT headers)
- [PE Section Headers] (for all section headers)
- [PE Image Headers] (for the "optional" header, only present in images)

This would be similar to your "option 1", except collapse `.text Header` and `.pdata Header` into `[PE Section Headers]`.

Rationale: each individual section header is not big enough to be interesting, but seeing them all grouped together under `[PE Section Headers]` can help users know how much file space in total they are spending for having N sections in their file.

src/pe.cc
```diff
@@ -12,56 +12,56 @@
...
+constexpr size_t kNameSize = 8;
+constexpr size_t kSymbol16Size = 18;
+constexpr size_t kSymbol32Size = 20;
+constexpr size_t kSectionSize = 40;
+constexpr size_t kRelocationSize = 10;
+constexpr size_t kBaseRelocationBlockSize = 8;
+constexpr size_t kImportDirectoryTableEntrySize = 20;
+constexpr size_t kResourceDirectoryTableSize = 16;
+constexpr size_t kResourceDirectoryEntriesSize = 8;
+constexpr size_t kResourceDataEntrySize = 16;
```
learn-more, line 39: @haberman I hope you don't mind, I changed these as well.

src/pe.cc
```diff
@@ -109,23 +110,37 @@ bool PeFile::Initialize() {
     return false;
   }
 
-  string_view exe_header =
-      GetRegion(dos_header_.AddressOfNewExeHeader, sizeof(pe_header));
-  memcpy(&pe_header_, exe_header.data(), exe_header.size());
+  PE_TYPE Magic;
+  auto pe_end =
+      CheckedAdd(dos_header_.AddressOfNewExeHeader, sizeof(pe_header_));
+  if (CheckedAdd(pe_end, sizeof(Magic)) > data_.size()) {
```
learn-more, line 116: @haberman I changed this code, because I also wanted to check the magic field of the optional header following the PE header.
If I understood the problem from fuzzing correctly this should still work fine?

Would it be possible for you to add a yaml file for the 'bad' PE file that the fuzzer produced?

src/pe.cc
```diff
@@ -109,23 +110,37 @@ bool PeFile::Initialize() {
...
-  // TODO(mj): Parse PE header further to determine this
-  is_64bit_ = false;
+  memcpy(&Magic, data_.data() + pe_end, sizeof(Magic));
+
+  if (Magic != PE_TYPE::PE32 && Magic != PE_TYPE::PE32_PLUS) {
+    // Unknown PE magic
+    return false;
+  }
+
+  is_64bit_ = Magic == PE_TYPE::PE32_PLUS;
```
learn-more, line 137: Some structs are different on 64 bit (none of those are used yet, but I wanted to add this check in advance)

learn-more (commented):
Some notes for your consideration.

src/pe.cc
haberman: Yep np!

src/pe.cc
haberman: Every time we fix a fuzzing bug, the payload that crashed the program is added to the [fuzz corpus](https://github.com/google/bloaty/tree/master/tests/testdata/fuzz_corpus). The [`fuzz_test` test](https://github.com/google/bloaty/blob/master/CMakeLists.txt#L382-L384) will always test against every payload in the corpus.

If the code is still passing `fuzz_test`, we should be good!

We don't specify these payloads with yaml, because often they aren't valid input files at all, so `yaml2obj` wouldn't be capable of producing them.

tests/PE/sections/empty-pe32-dll.test
```diff
@@ -0,0 +1,90 @@
+# RUN: %yaml2obj %s -o %t.obj
+# RUN: %bloaty --raw-map %t.obj | %FileCheck %s --dump-input fail
+
+--- !COFF
+OptionalHeader:
```
haberman, line 5: How is this optional header reflected in the map?

I was hoping to report these as `[PE Image Headers]` or similar.

src/pe.cc
learn-more: Great, thanks.

tests/PE/sections/empty-pe32-dll.test
learn-more: It is included in the `[PE Headers]` entry.
For a PE file there are always:
- DOS Header
- Optional DOS stub (between DOS and NT header)
- NT Headers (See IMAGE_OPTIONAL_HEADER32 and IMAGE_OPTIONAL_HEADER64 in winnt.h):
  - Signature
  - FileHeader
  - OptionalHeader (contradicting it's name, it is not optional)

learn-more: @haberman is that good enough, or do you want me to rename `[PE Headers]` to `[PE Image Headers]`?

haberman: > OptionalHeader (contradicting it's name, it is not optional)

I don't think that's entirely true: I believe it's required on DLLs and executables, but not on object files. It's the equivalent of what ELF calls "program headers."

In my comment above I was thinking of splitting the optional headers out into something like `[Image Headers]`. But since PE is having a lot more header types overall, it might not be worth splitting this out. You've already split out the section headers, which is the most important part.

So I think this is great as is.

PR #267, opened 2021-08-04
Split and updated the docs.

haberman:


haberman:
Preview:

- https://github.com/haberman/bloaty/blob/docs/README.md
- https://github.com/haberman/bloaty/blob/docs/doc/using.md
- https://github.com/haberman/bloaty/blob/docs/tests/README.md

compnerd (approved):

README.md
```diff
@@ -184,826 +35,59 @@ $ ./bloaty bloaty -d compileunits
...
-   1.0%  45.2Ki   2.4%  12.4Ki    numbers.cc.o
-   0.9%  40.9Ki   2.2%  11.4Ki    charconv.cc.o
-   0.9%  38.8Ki   1.2%  6.10Ki    int128.cc.o
-   0.7%  30.1Ki   1.1%  5.58Ki    str_cat.cc.o
-   0.6%  24.1Ki   0.8%  3.92Ki    string_view.cc.o
-   0.5%  21.2Ki   0.6%  3.21Ki    throw_delegate.cc.o
-   0.4%  19.2Ki   0.4%  2.26Ki    ascii.cc.o
- 100.0%  4.23Mi 100.0%   512Ki    TOTAL
+$ cmake -G Ninja .
+$ ninja
```
compnerd, line 69:
```suggestion
$ cmake -B build -G Ninja -S .
$ ninja -C build
$ ninja -C build -t install
```

This ensures that the copy-paste will run out-of-tree builds, and uses the `install` target to generate a generally unix-like install image.

haberman: `ninja -C build -t install` fails on my machine:

```
ninja: fatal: unknown tool 'install'
```

Also should we use a cmake invocation for the second two commands so it will work even if the user doesn't decide to use `ninja`?

compnerd: Updated the suggestion to use `cmake` to drive the build tool, that should make it easier to follow/update and fix the error.

PR #286, opened 2021-09-08
Support DW_FORM_strx3

tengyifei:
c.f. http://www.dwarfstd.org/doc/DWARF5.pdf page 218-219, DW_FORM_strx3 implies a 3-byte integer.

I have a few test ELF files that contain `DW_FORM_strx3` but unfortunately their size is in the 50 megabytes. I'm not sure of a better way to test this.


haberman:
If you are feeling ambitious, you could add a `yaml2obj` test in https://github.com/google/bloaty/tree/master/tests/dwarf/debug_info. See the existing tests there for examples.

However this is such a straightforward extension of what's already there that I am happy to merge without a specific test for it.

PR #313, opened 2022-07-11
Preserve Size Information During Diff

brandonvu12:
The changes enable the CSV output to provide the old and new vm/file size during diff mode. The act of subtracting the two binaries was moved to a later stage.


src/bloaty.h
```diff
@@ -337,18 +337,31 @@ void DisassembleFindReferences(const DisassemblyInfo& info, RangeSink* sink);
...
   double vmpercent;
   double filepercent;
+
+  // The size of this row in a diff base.  Same as `size` for non-diff.
+  DomainSizes old_size = {0, 0};
+  // The current size of this row (this is not the same as `size` in the case
+  // of a diff -- `size` is the delta).
+  DomainSizes new_size = {0, 0};
+
+  DomainSizes capacity = {0, 0};
```
haberman, line 363: I think we're leaving out capacity for now?

src/bloaty.h
```diff
@@ -337,18 +337,31 @@ void DisassembleFindReferences(const DisassemblyInfo& info, RangeSink* sink);
...
+  DomainSizes size = {0, 0};
+  DomainSizes filtered_size = {0, 0};
+
   int64_t other_count = 0;
   int64_t sortkey;
   double vmpercent;
   double filepercent;
+
+  // The size of this row in a diff base.  Same as `size` for non-diff.
+  DomainSizes old_size = {0, 0};
```
haberman, line 358: What if we stored only `old_size` or only `new_size`, and let the other get derived from that?

I wonder if it might help make the code a little simpler.

haberman (changes requested):

src/bloaty.h
haberman: In particular, you wouldn't need to fill in either for a non-diff run.

brandonvu12: Storing only one struct and deriving the other by adding the `size` would be done during printing as we will not have a struct to hold those values. I am not sure, but should we refrain from doing any kind of logic or calculations within a printing function? Or is there another way of obtaining those sizes?

src/bloaty.cc
```diff
@@ -845,6 +879,10 @@ void RollupOutput::PrintToCSV(std::ostream* out, bool tabs) const {
   std::vector<std::string> names(source_names_);
   names.push_back("vmsize");
   names.push_back("filesize");
+  names.push_back("original_vmsize");
```
haberman, line 882: Can we add these extra columns only in diff mode?

PR #315, opened 2022-08-23
Provide size report despite not translating VM pointer

brandonvu12:
Size reporting for certain binaries are getting a "Can't translate VM pointer" error and as a result, not outputting size reports.


brandonvu12:
fix #315

haberman (approved):

PR #326, opened 2022-11-07
third_party: update googletest to current main

compnerd:
Update the pinned revision for googletest to pick up google/googletest#2815 which should allow us to repair the macOS builds.


haberman:
Build still appears broken.

compnerd:
This failure seems different from the previous one(s).  I think that there is a missing dependency somewhere.

PR #327, opened 2022-11-07
build: add a missing dependency on capstone for libbloaty

compnerd:
`libbloaty` depends on capstone (at least in `disassembler.cc`).  Use `target_link_libraries` to indicate that `libbloaty` should link against capstone which will also ensure that we have the proper header search path for the capstone headers and that capstone is a build dependency for `libbloaty`.  This should hopefully avoid a latent build race that was uncovered in #325.


haberman:
This appears to cause a build error on Windows.

compnerd:
This was an incorrect analysis, and the workaround for this is squashed into #326

PR #354, opened 2023-08-09
[WebAssembly] Update yaml2obj results

aheejin:
After
https://github.com/llvm/llvm-project/commit/1b21067cf247c62c2442daa7ee2d3915249d1ee2, if sections in Wasm YAML files don't have `HeaderSecSizeEncodingLen` field specified, yaml2obj will automatically pad header section to 5 bytes, which was a byproduct of fixing some llvm-objcopy issue.

cc @dschuff

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: llvm/llvm-project
-          ref: llvmorg-13.0.0-rc1
+          ref: llvmorg-17.0.0-rc3
           path: llvm-project
           fetch-depth: 1
       - name: Get LLVM Revision
@@ -71,7 +71,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           repository: llvm/llvm-project
-          ref: llvmorg-13.0.0-rc1
+          ref: llvmorg-17.0.0-rc3
           path: llvm-project
           fetch-depth: 1
       - name: Get LLVM Revision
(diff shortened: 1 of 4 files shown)


haberman:
Looks like this is failing in CI.

aheejin:
Does this repo use tot LLVM?

haberman:
Yes, but we are pinned to a specific version, see: https://github.com/google/bloaty/blob/2757d3ec0fec07f4ac7f2f96fe8f0b020c2d84fb/.github/workflows/build.yml#L25

Perhaps we need to update that version to something newer?

aheejin:
Yeah maybe we should. But the yaml2obj-changing commit has not been included in any of release candidates so far because it is so recent, so I opened https://github.com/llvm/llvm-project/issues/64720 to backport it to the next release, so I have to wait until it is merged. Reason I'd like to merge this is Google's internal toolchain's bloaty uses tot LLVM so it is breaking. (I think you're cc'ed in that internal bug too.)

haberman:
I think it could also be possible to check out llvm at an arbitrary commit if that helps.

aheejin:
Ah I didn't know that. Then should we do that? I'll change the ref to https://github.com/llvm/llvm-project/commit/1b21067cf247c62c2442daa7ee2d3915249d1ee2 then, but it also looks `llvmorg-17.0.0-rc3` will be tagged Aug 22nd, which is not far. Maybe we can update it to a released version later again.

aheejin:
Can you enable the CI? Thanks!

PR #358, opened 2023-08-22
Update LLVM to 17.0.0-rc3 in CI

aheejin:
This is necessary to make #354's CI pass, but I thought it would be good to submit this standalone to see everything works fine with this before doing that.


aheejin:
Ah, right, this is supposed to fail because Wasm test case updates pending in #354 are not included here....
https://github.com/google/bloaty/actions/runs/5944362753/job/16123263190?pr=358

Will just include the update there as well, and will close this.

PR #365, opened 2023-10-31
add_range: Replace THROW with WARN for early return

erick-pfeifer:
Replaces THROW statement from the RangeSink::AddRange function, to WARN, when the translator CoversRange test fails. This will allow for the operation to resume, while ignoring those failing ranges, instead of throwing an exception.

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -1334,7 +1334,8 @@ void RangeSink::AddRange(const char* analyzer, string_view name,
   if (translator_) {
     if (!translator_->vm_map.CoversRange(vmaddr, vmsize) ||
         !translator_->file_map.CoversRange(fileoff, filesize)) {
-      THROW("Tried to add range that is not covered by base map.");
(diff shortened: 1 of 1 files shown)


src/bloaty.cc
```diff
@@ -1334,7 +1334,8 @@ void RangeSink::AddRange(const char* analyzer, string_view name,
   if (translator_) {
     if (!translator_->vm_map.CoversRange(vmaddr, vmsize) ||
         !translator_->file_map.CoversRange(fileoff, filesize)) {
-      THROW("Tried to add range that is not covered by base map.");
+      WARN("Tried to add range not covered by base map, range ignored.");
```
haberman, line 1337: Please add all the function parameters to the debugging output (analyzer, name, etc)

erick-pfeifer: Done, thank you for pointing that out.

haberman (approved):

PR #366, opened 2023-11-18
Support omission of exception header

planetmarshall:
Fix detection of `DW_EH_PE_omit` values in the exception header


EricRahm:
Thanks for the contribution @planetmarshall! I'm going to rebase this and add a test, if all checks look happy we'll get it landed.

Issue #378, opened 2024-07-12
bloaty shows 'region out-of-bounds' error on ELF files with a 0-sized segment at an offset greater than the ELF file size 


kjteske:
`bloaty` errors processing an ELF file where a 0-sized segment is not backed by the ELF file's contents:

```
$ bloaty <binary>
bloaty: region out-of-bounds  
```

This happens when `bloaty` is iterating segments and gets to the "TLS" segment in this particular file. `readelf` shows the `.tbss` section / TLS segment is at address `0x00690000`, with `MemSiz` of 4 bytes but `FileSiz` is 0 bytes. With some irrelevant details redacted:

```
$ readelf -l -S <binary>
Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
   ...
  [20] .tbss             NOBITS          00690000 680000 000004 00 WAT  0   0  4
   ...

Program Headers:
  Type           Offset   VirtAddr   PhysAddr   FileSiz MemSiz  Flg Align
  EXIDX          0x648a04 0x00658a04 0x00658a04 0x36da8 0x36da8 R   0x4
  ...
  TLS            0x680000 0x00690000 0x00690000 0x00000 0x00004 R   0x4
  GNU_STACK      0x000000 0x00000000 0x00000000 0x00000 0x00000 RW  0x10

 Section to Segment mapping:
  Segment Sections...
   00     .ARM.exidx 
   ...
   08     .tbss 
   09  
```

Since `FileSiz` is 0 bytes, it doesn't need to be backed by the file. And we see the file indeed is smaller than the `0x00690000` offset:
```
$ printf "0x%x\n" $(stat --format "%s" <binary>)
0x68aaf0
```

The crash is within `ElfFile::ReadSegment()` when ` header->p_filesz` is `0`. When `p_filesz` is `0`, we should skip calling `GetRegion()`.

haberman:
Sorry for the slow reply.

I'm not quite understanding why this made Bloaty crash. Your report seems to be stating that `p_offset` is `0x00690000`, but from my reading of your `readelf` output, `p_offset=0x680000`, which is less than your file size `0x68aaf0`.

[...]

PR #379, opened 2024-07-12
Fix bloaty error if zero-sized segment is at offset larger than file size

kjteske:
https://github.com/google/bloaty/issues/378


kjteske:
I can add a test if this fix looks good.

kjteske:
@haberman - any thoughts on this PR? I can look into adding a test if this makes sense and looks good.

PR #387, opened 2024-08-30
Wasm source maps

mason-lgtm:
Add support for "inlines" and "compileunits" sources for Wasm using source maps.

Adds a new option "--source-map" which can be specified to map a URL to source map file:

`--source-map=http://myurl/output.wasm.map=output.wasm.map`

The source map URL is found in the Wasm binary in the `sourceMappingURL` section.

We also support specifying the source map by file name, which is useful for specifying multiple source maps in diff mode where the `sourceMappingURL` might be identical between the binaries.

`--source-map=path/to/output.wasm=output.wasm.map`

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -1973,6 +1973,11 @@ Options:
   -c FILE            Load configuration from <file>.
   -d SOURCE,SOURCE   Comma-separated list of sources to scan.
   --debug-file=FILE  Use this file for debug symbols and/or symbol table.
+  --source-map=ID=FILE
+                     Use this source map file for the binary. The ID can be
+                     the build ID (or Wasm sourceMappingURL) or the file path
+                     as specified in the command line.
+                     Currently only supported for Wasm.
   -C MODE            How to demangle symbols.  Possible values are:
   --demangle=MODE      --demangle=none   no demangling, print raw symbols
                        --demangle=short  demangle, but omit arg/return types
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -49,18 +49,18 @@ typedef size_t z_size_t;
 #include <sys/stat.h>
 #include <sys/types.h>
 
-#include "absl/debugging/internal/demangle.h"
-#include "absl/memory/memory.h"
-#include "absl/strings/numbers.h"
-#include "absl/strings/str_join.h"
-#include "absl/strings/string_view.h"
-#include "absl/strings/substitute.h"
-#include "bloaty.h"
-#include "bloaty.pb.h"
-#include "google/protobuf/io/zero_copy_stream_impl.h"
-#include "google/protobuf/text_format.h"
-#include "re.h"
-#include "util.h"
+#include "third_party/absl/debugging/internal/demangle.h"
+#include "third_party/absl/memory/memory.h"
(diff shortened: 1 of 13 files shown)


haberman (changes requested):

src/webassembly.cc
```diff
@@ -412,8 +427,20 @@ class WebAssemblyObjectFile : public ObjectFile {
...
+    // Use the sourceMappingURL as the build ID to be able to match to the
+    // source map.
+    std::string id;
+
+    FindSection(file_data().data(), "sourceMappingURL",
+                [&id](Section& section) {
+                  uint32_t size = ReadVarUInt32(&section.contents);
+                  string_view source_mapping_url =
+                      ReadPiece(size, &section.contents);
+                  id.resize(size);
```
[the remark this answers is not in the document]
mason-lgtm: Thanks! done

gkdn:
Hi. Wanted to quickly check if there is additional work needed for PR or if it is ready to go?

src/bloaty.cc
```diff
@@ -2166,6 +2196,8 @@ bool DoParseOptions(bool skip_unknown, int* argc, char** argv[],
       }
     } else if (args.TryParseOption("--source-filter", &option)) {
       options->set_source_filter(std::string(option));
+    } else if (args.TryParseOption("--source-map", &option)) {
```
haberman, line 2199: Please add documentation for this flag to the usage string and into the user guide.

tests/wasm/sourcemap_compileunits.test
```diff
@@ -0,0 +1,30 @@
+# RUN: %yaml2obj %s -o %t.obj
```
haberman, line 1: Thanks so much for adding yaml2obj tests. Could we add one for diff mode, so we can test our design for distinguishing between different versions of the same file?

mason-lgtm: Thanks. I think the simplest way to do this would be to make a copy of the Wasm binary generated by yaml2obj, then create 2 different source maps for each of these. Made the change. PTAL

src/bloaty.cc
mason-lgtm: Done - added usage notes to using.md and added the option to the usage string

jordo:
I'm trying to use this with an emscripten build, and using the generated source map file:

Example:
```
godot.web.template_release.wasm32.nothreads.wasm
godot.web.template_release.wasm32.nothreads.wasm.map
```

Passing the map file as an argument doesn't seem to work:

```
bloaty godot.web.template_release.wasm32.nothreads.wasm -d compileunits --source-map ./godot.web.template_release.wasm32.nothreads.wasm.map
bloaty: Source map filename './godot.web.template_release.wasm32.nothreads.wasm.map' must have a build id and file name separated by '='
```

It's not clear to me what the cmd line help actually means:

```
  --source-map=ID=FILE
                     Use this source map file for the binary. The ID can be
                     the build ID (or Wasm sourceMappingURL) or the file path
                     as specified in the command line.
                     Currently only supported for Wasm.
```

haberman:
There is a bit more documentation here: https://github.com/google/bloaty/blob/main/doc/using.md#source-maps

Issue #360, opened 2023-09-06
zlib build failure when using bundled version


carlosperate:
Builds fine when `libz-dev` is installed in Ubuntu 22.04, but without it, it tries to use the local submodule sources and fails to find `zlib.h`.

```
$ cmake -B build -G Ninja -S . && cmake --build build
-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE)
-- Could NOT find ZLIB (missing: ZLIB_LIBRARY ZLIB_INCLUDE_DIR)
-- Checking for one of the modules 're2'
-- Checking for one of the modules 'capstone'
-- Checking for one of the modules 'protobuf'
-- System re2 not found, using bundled version
-- System capstone not found, using bundled version
-- System protobuf not found, using bundled version
-- System zlib not found, using bundled version
-- Setting build type to 'RelWithDebInfo' as none was specified.
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
CMake Deprecation Warning at third_party/capstone/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


CMake Warning (dev) at third_party/capstone/CMakeLists.txt:2 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    PROJECT_VERSION
    PROJECT_VERSION_MAJOR
    PROJECT_VERSION_MINOR
[...]

PR #388, opened 2024-09-12
Fixes to build with bundled zlib

peter23:
Fixes `'zconf.h' file not found` and `zlib.h file not found`.

Should fix https://github.com/google/bloaty/issues/360


haberman (approved):

EricRahm:
We just landed a similar change #432, thanks for identifying this fix!

PR #390, opened 2024-12-01
Fix crash in StrictSubstr on `-ohos` artifacts

jschwe:
Fixes a "region out-of-bounds" assertion failure in StrictSubstr when running bloaty on a Rust shared library compiled for `aarch64-unknown-linux-ohos`.

When `remaining_` is just the "\0" character, we are done and should return.

### DIFF
--- a/src/elf.cc
+++ b/src/elf.cc
@@ -416,7 +416,7 @@ void ElfFile::Section::ReadRelocationWithAddend(Elf64_Word index,
 }
 
 void ElfFile::NoteIter::Next() {
-  if (remaining_.empty()) {
+  if (remaining_.empty() || remaining_.compare("\0")) {
     done_ = true;
     return;
   }


EricRahm:
@jschwe Thanks for the proposed fix! I rebased and tested this and it looks like it regresses a few tests:
```
Failed Tests (5):
  bloaty :: dwarf/debug_info/dw_form_line_strp.test
  bloaty :: dwarf/debug_info/ignore-declarations.test
  bloaty :: dwarf/debug_info/pcpair-incomplete.test
  bloaty :: dwarf/debug_info/pcpair.test
  bloaty :: dwarf/range_lists/base-addr-selection.test
```
I'm guessing the fix is a bit more nuanced. It might be worth filing an issue including a binary that reproduces to track this.

If you're up for it we'd want to add a `lit` test to reproduce the issue and a rebased and updated fix.

PR #391, opened 2025-01-17
improve handling of third-party dependencies in CMake

marzer:
Adds two additional `BLOATY_PREFER_SYSTEM_*` options to control how protobuf and re2 are sourced. Defauts are set to `YES` to preserve existing workflows.

Also lazy-initializes git submodules only when they are required. Previously the build was always cloning submodules even if the system install was selected, which slowed the CMake configure step down quite a bit in some cases.


marzer:
Hello? _Is this thing on?_

EricRahm:
@marzer Thanks for the cleanup, I just hit this pain point myself! Would you mind rebasing on tip-of-tree, we just landed a few changes in `CMakeLists.txt` that conflict with this branch. Once that's up I'll trigger our CI and we should be good to go.

marzer:
Sure thing, will do.

marzer:
Update: done @EricRahm 😄

CMakeLists.txt
EricRahm: I think we just missed an `init_submodule` here, I was able to repro the build failure by running:
```
git submodule deinit --all -f
cmake -S . -B build/cmake_deps -G Ninja
```

EricRahm (changes requested):
Looks like we've got a build failure, I suspect it's just a minor fix.

EricRahm:
Looks like more issues with the windows config.

marzer:
@EricRahm 

> Looks like more issues with the windows config.

Well, this is a head-scratcher. The following works perfectly in a windows CMD terminal:

```
git submodule update --init --recursive third_party/capstone
```

But when the _exact_ same command is invoked by my `init_submodule()` function within CMake, it silently fails, same for a few other submodules also. Only  `abseil-cpp`, `re2` and `googletest` are OK, and it's not obvious to me how they are different to the others. Any ideas?

I've tried completely removing and re-adding one of the affected submodules, but that did not help.

One possible, if extreme, solution: abandon submodules entirely, in favour of FetchContent or similar.

EricRahm:
> @EricRahm
> 
> > Looks like more issues with the windows config.
> 
> Well, this is a head-scratcher. The following works perfectly in a windows CMD terminal:
> 
> ```
> git submodule update --init --recursive third_party/capstone
> ```
> 
> But when the _exact_ same command is invoked by my `init_submodule()` function within CMake, it silently fails, same for a few other submodules also. Only `abseil-cpp`, `re2` and `googletest` are OK, and it's not obvious to me how they are different to the others. Any ideas?

I think maybe init_submodule isn't called at all? I was able to repro with:
```
git submodule deinit --all -f
cmake ...
```

My guess is we just need to add explicit `init_submodule` calls to the windows path of the config.

marzer:
@EricRahm Yep, that was it. I somehow completely missed the `if(UNIX OR MINGW)` condition 😅 Have pushed a fix.

EricRahm (approved):

EricRahm:
Looks good, thanks for the contribution @marzer !

Issue #392, opened 2025-01-18
Crash when debug info parameter is used


vogel76:
bloaty was used with --debug-file option. Unfortunetely it crashes.

bloaty -d compileunits --source-map=wax.common.wasm.map=./wasm/build_wasm/wax.common.wasm.map ./wasm/build_wasm/wax.common.wasm --demangle=full --domain=file -n 0 --debug-file=./wasm/build_wasm/wax.common.wasm.debug.wasm

EricRahm:
Fixed in #393

PR #393, opened 2025-01-18
Issue #392 Fixed crash when debuginfo file was passed to the tool.

vogel76:
I assumed that assertion condition should refer passed argument, not class member before assignment.


vogel76:
I assume windows build problems are unrelated to such changes. Please let me know if I should do anything more in this subject.

EricRahm (approved):
Nice catch, thanks for the fix!

PR #399, opened 2025-04-05
absl::string_view -> std::string_view

haberman:
Now that we depend on C++17, there is no need to use the
ABSL version.


dlech:
This breaks building with system ABSL on Ubuntu 24.04 which has `#define ABSL_OPTION_USE_STD_STRING_VIEW 0` in `/usr/include/absl/base/options.h`.

PR #408, opened 2025-08-24
Fix a typo in the DWARF attr parser

FlorentRevest:
I'm not sure if this actually affects any of the Bloaty features but I was reusing the DWARF parsing code for a separate project and noticed something was off there.


EricRahm (approved):
lgtm, thanks for the fix!

PR #410, opened 2025-10-06
Fix build with cmake 4.0+.

### LINKED ISSUE #400: Fails to build with cmake 4.0
Using cmake 4.0.0 the build fails:

```
-- Found Threads: TRUE
CMake Deprecation Warning at third_party/re2/CMakeLists.txt:6 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


CMake Error at third_party/capstone/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

rjmansfield:
Updates third party dependencies which require updates for cmake 4+

Update to capstone 5.0.6.
Update to protobuf 31.1.
Update to googletest 1.17.0.
Update to abseil-cpp 20250814.1.

Also fixes issue where system abseil-cpp headers were not included as a dependency when building libbloaty.

Fixes #400


EricRahm:
@rjmansfield Thanks for the fixes! We're interested in updating our dependencies-- it'll fix our CI builder for macOS which would be awesome.  I triggered a CI run and there are some failures/improvements:
- It looks like macOS is building again
- The fuzzers are failing, I've fixed that at tip-of-tree, if you merge that into your branch we can respin the build to confirm
- The windows builders are failing, this seems to be from your build file changes (or missing a few windows branches)

Are the windows builds something you can follow up on?

rjmansfield:
OK, sounds good. I have some other fixes for macho that I'd like to contribute as well. I don't have a windows machine/vm super handy.  I'll review first and try to spot what I missed. If not, I'll set one up.

EricRahm:
> OK, sounds good. I have some other fixes for macho that I'd like to contribute as well.

Happy to help get those landed as well, getting our CI builder for macOS working again is a great first step.

> I don't have a windows machine/vm super handy. I'll review first and try to spot what I missed. If not, I'll set one up.

Let me know when you have a potential fix and I'll spin up the builder, I can also look into setting up a local windows build and work with you on getting this landed.

rjmansfield:
> Let me know when you have a potential fix and I'll spin up the builder, I can also look into setting up a local windows build and work with you on getting this landed.

I think I spotted the issue. I missed the change needed to  add_subdirectory(third_party/protobuf/cmake) in the windows case

EricRahm:
> I think I spotted the issue. I missed the change needed to add_subdirectory(third_party/protobuf/cmake) in the windows case

Looks better! Maybe just one more issue?

> CMake Error at CMakeLists.txt:155 (set_property):
  set_property could not find TARGET capstone-static.  Perhaps it has not yet
  been created.

rjmansfield:
OK, looks like I also missed the capstone library change as well for windows.  I'm hoping the latest will fix it.   Sorry we're iterating this way. If this is too annoying, please let me know and I'll just set up a machine.

rjmansfield:
Well, windows x64 built but win32 had an ICE. :(

`D:\a\bloaty\bloaty\bloaty\third_party\protobuf\src\google/protobuf/descriptor_visitor.h(108,1): fatal  error C1001: Internal compiler error. [D:\a\bloaty\bloaty\build\bloaty\third_party\protobuf\protoc-gen-upbdefs.vcxproj]`

EricRahm:
> Well, windows x64 built but win32 had an ICE. :(
> 
> `D:\a\bloaty\bloaty\bloaty\third_party\protobuf\src\google/protobuf/descriptor_visitor.h(108,1): fatal error C1001: Internal compiler error. [D:\a\bloaty\bloaty\build\bloaty\third_party\protobuf\protoc-gen-upbdefs.vcxproj]`

That's...impressive. I retriggered it to see if it's consistent, either way we probably don't want to land a flake.

I'm not sure the best next step here, file a bug against protobuf and then maybe we could try an older version of protobuf for now? I'm going to see if I can get a windows machine building bloaty to repro locally.

rjmansfield:
Looks like others have hit this as well. 

https://developercommunity.visualstudio.com/t/Compiler-internal-error-when-compiling-l/10907298?sort=newest
https://developercommunity.visualstudio.com/t/fatal--error-C1001:-Internal-compiler-er/10906076

In https://github.com/grpc/grpc/issues/39819 they worked around it by using C++ 20 instead of C++17

haberman:
Hmm, protobuf's CI has a win32 test that seems to be passing: https://github.com/protocolbuffers/protobuf/actions/runs/18794895630/job/53633406124

I wonder if we can somehow make Bloaty's CI more closely resemble Protobuf's, which I think is here: https://github.com/protocolbuffers/protobuf/blob/b5b9f2c0ba359fc54cde41baf129ce561a41d43c/.github/workflows/test_cpp.yml#L476-L483

EricRahm:
> Hmm, protobuf's CI has a win32 test that seems to be passing: https://github.com/protocolbuffers/protobuf/actions/runs/18794895630/job/53633406124
> 
> I wonder if we can somehow make Bloaty's CI more closely resemble Protobuf's, which I think is here: https://github.com/protocolbuffers/protobuf/blob/b5b9f2c0ba359fc54cde41baf129ce561a41d43c/.github/workflows/test_cpp.yml#L476-L483

They're targeting cxx17 but disabling some configs, I wonder if that would help:
```
cmake . -DCMAKE_CXX_STANDARD=17 -G Ninja \
  -Dprotobuf_WITH_ZLIB=OFF \
  -Dprotobuf_BUILD_CONFORMANCE=OFF
  -DCMAKE_C_COMPILER_LAUNCHER=sccache \
  -DCMAKE_CXX_COMPILER_LAUNCHER=sccache \
  -Dprotobuf_ALLOW_CCACHE=ON
```

haberman:
I don't think that any of those settings would necessarily prevent `descriptor_visitor.h` from being compiled. That header looks to be a part of the main protobuf library, so disabling zlib, onformance, etc. won't change that afaict (FWIW, I am on the Protobuf team).

I was wondering if these settings from Protobuf might make a difference though:
```yaml
            os: windows-2022
            vsversion: '2022'
            windows-arch: 'win32'
```

Protobuf also seems to use https://github.com/ilammy/msvc-dev-cmd to set up MSVC: https://github.com/protocolbuffers/protobuf/blob/b5b9f2c0ba359fc54cde41baf129ce561a41d43c/.github/workflows/test_cpp.yml#L511-L517

It seems worth trying to replicate the Protobuf setup. I would expect that to fix the compiler error.

EricRahm:
Alright, I've got a windows machine setup and confirmed locally that specifying C++ 20 does fix things (and I confirmed the ICE is reproducible). This seems like a reasonable approach, or maybe just an interim fix. Chasing down the github config might be a bit more work given protobuf's bespoke environment (they're using ninja, a different shell, etc).

haberman:
We could go the way of C++20, but we have to make sure that no actual use of C++20 creeps into the code base.

I guess this will get tested anyway since other platforms are building as C++17. So that SGTM.

rjmansfield:
I pushed an update that limits the workaround to forcing C++20 only on the protobuf targets when building with MSVC.  Hopefully that'll work and limit the scope of the work around.

https://github.com/google/bloaty/pull/410/commits/112c4e10f4db7e6098bc02ea1ff7f3d62346a160#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR164

EricRahm:
Thanks for helping fix our builds @rjmansfield! The latest run looks good, lets get this landed.

PR #413, opened 2025-10-24
Add support for tags in import section.

davits:
Bloaty fails with "Unrecognized import kind: 4" when there is a tag in import section.

Minimum example:
```
cat > example.cpp <<EOF
#include <emscripten.h>
EMSCRIPTEN_KEEPALIVE void foo() { throw "error"; }
EOF
em++ example.cpp -sMAIN_MODULE=2 -fwasm-exceptions -o example.wasm
bloaty -d symbols example.wasm
```

### DIFF
--- a/src/webassembly.cc
+++ b/src/webassembly.cc
@@ -163,6 +163,7 @@ struct ExternalKind {
     kTable = 1,
     kMemory = 2,
     kGlobal = 3,
+    kTag = 4,
   };
 };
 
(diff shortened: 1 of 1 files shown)


EricRahm:
@davits Thanks for the fix! Would you mind rebasing on top of the tip-of-tree? We recently fixed some CI issues and I'd like to make sure our builds and tests are passing.

Also, if you're up for it it'd be great to add a test for this. The tests are a little tricky to setup but there are some [good examples for wasm specifically](https://github.com/google/bloaty/blob/6dcd80bc4d07c6ce1c095fc94a705f7a70e283fc/tests/wasm/sections.test).

PR #417, opened 2025-10-27
Fix macOS builder

### LINKED ISSUE #416: macOS github builder is broken
The `macOS` github build actions are failing due to some sort of cmake version issue in the `third_party/capstone` dependency:

```
CMake Error at third_party/capstone/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
-- Configuring incomplete, errors occurred!

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```
For example: https://github.com/google/bloaty/actions/runs/18853362004/job/53795056105?pr=364

EricRahm:
Workaround a failure to build the project on the github `macOS`
builders. This does two main things:
  1. Switches over to using a prebuilt `capstone` library to work
     around various build errors.
  2. Sets `CMAKE_POLICY_VERSION_MINIMUM=3.5` to fix protobuf build errors

`brew install capstone` is used, but this sets an invalid include path
so additionally the `pkg-config` file for capstone is patched to update to
the proper path.

Fixes #416

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -68,6 +68,18 @@ jobs:
     steps:
       - uses: seanmiddleditch/gha-setup-ninja@master
 
+      - name: Update Homebrew
+        run: brew update
+
+      - name: Install capstone
+        run: |
+          brew install capstone
+          if [ -d /opt/homebrew ]; then HB=/opt/homebrew; else HB=/usr/local; fi
+          CAPSTONE_PREFIX=$(brew --prefix capstone)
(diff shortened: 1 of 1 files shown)


EricRahm:
FWIW it looks like #410 would fix this in a proper way (upgrading the capstone and proto deps).

EricRahm:
#410 took care of this.

PR #419, opened 2025-10-28
Handle truncated Mach-O debug section names

rjmansfield:
Mach-O section names are limited to 16 characters. When DWARF section names exceed this limit, they get truncated. The section name __debug_str_offsets is the only standard DWARF
section that exceeds this limit, and gets truncated to __debug_str_offs.

After the Mach-O parser strips the __debug_ prefix, GetFieldByName() receives "str_offs" instead of the expected "str_offsets", causing the debug_str_offsets section to not be recognized.

### DIFF
--- a/src/dwarf.cc
+++ b/src/dwarf.cc
@@ -200,7 +200,11 @@ string_view* File::GetFieldByName(string_view name) {
     return &debug_addr;
   } else if (name == "str") {
     return &debug_str;
-  } else if (name == "str_offsets") {
+  } else if (name == "str_offsets" || name == "str_offs") {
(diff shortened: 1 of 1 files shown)


EricRahm:
@rjmansfield would you mind rebasing this PR (and your others) off of the latest main? That'll allow us to trigger a (hopefully) passing CI run.

rjmansfield:
OK, the current patches should be based on main now.

EricRahm (approved):
Thanks for the fix, the detailed comment was particularly helpful! Once CI passes we'll get this landed.

PR #420, opened 2025-10-28
Do two-pass attribute parsing in ReadTopLevelDIE() 

rjmansfield:
This ensures base addresses are set before parsing attributes that depend on them.

In DWARF v5, DW_AT_name can use the DW_FORM_strx form, which requires DW_AT_str_offsets_base to be parsed first. Previously, all attributes were parsed in a single pass, which meant that if DW_AT_name appeared before DW_AT_str_offsets_base in the DIE, parsing would fail with "premature EOF reading variable-length DWARF data".


EricRahm (approved):
lgtm, thanks for the clear comments and the included tests!

PR #423, opened 2025-10-29
Fix DWO file path handling for absolute paths

rjmansfield:
Previously, DWO file paths were always constructed by concatenating comp_dir and dwo_name, which failed when dwo_name was already an absolute path. This resulted in malformed paths like "/comp/dir//abs/path".

### DIFF
--- a/src/dwarf.cc
+++ b/src/dwarf.cc
@@ -16,6 +16,7 @@
 #include <stdio.h>
 
 #include <algorithm>
+#include <filesystem>
 #include <initializer_list>
 #include <iostream>
 #include <limits>
@@ -591,6 +592,20 @@ struct DwoFilePointer {
   std::string dwo_name;
 };
 
+static std::string ConstructDwoPath(const DwoFilePointer& dwo_info) {
+  if (dwo_info.dwo_name.empty()) {
+    return "";
+  }
+
+  std::filesystem::path dwo_path(dwo_info.dwo_name);
+  if (dwo_path.is_absolute()) {
+    return dwo_info.dwo_name;
+  }
+
+  std::filesystem::path result = std::filesystem::path(dwo_info.comp_dir) / dwo_info.dwo_name;
+  return result.string();
+}
+
 // The DWARF debug info can help us get compileunits info.  DIEs for compilation
 // units, functions, and global variables often have attributes that will
 // resolve to addresses.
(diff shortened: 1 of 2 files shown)


rjmansfield:
It looks like I overlooked windows and the new test failed on the absolute path test with the drive letter. However this test illustrates the issue I ran into with the concatenation. e.g.

`couldn't open file 'ty/build/bloaty/tests/dwarf/debug_info/Output/gnu-split-dwarf-absolute-path.test.tmp/absolute.dwo/D:/a/bloaty/bloaty/build/bloaty/tests/dwarf/debug_info/Output/gnu-split-dwarf-absolute-path.test.tmp/absolute.dwo': 3bloaty: Failed to open file.`

I've updated the patch to use std::filesystem::path::is_absolute() instead which should be a more portable solution.

src/dwarf.cc
```diff
@@ -591,6 +592,25 @@ struct DwoFilePointer {
...
+static std::string ConstructDwoPath(const DwoFilePointer& dwo_info) {
+  if (dwo_info.dwo_name.empty()) {
+    return "";
+  }
+
+  std::filesystem::path dwo_path(dwo_info.dwo_name);
+  if (dwo_path.is_absolute()) {
+    // Absolute path
+    return dwo_info.dwo_name;
+  } else if (!dwo_info.comp_dir.empty()) {
```
EricRahm: Very minor nit, I don't think we need to check if the `comp_dir` is empty, path should do the right thing.

src/dwarf.cc
```diff
@@ -626,11 +646,14 @@ static void ReadDWARFDebugInfo(dwarf::InfoReader& reader,
           }
         });
 
-    if (!dwo_info.comp_dir.empty() && !dwo_info.dwo_name.empty()) {
-      auto file = MmapInputFileFactory().OpenFile(dwo_info.comp_dir + "/" + dwo_info.dwo_name);
-      dwarf::File dwo_dwarf;
-      cu.dwarf().open(*file, &dwo_dwarf, sink);
-      ReadDWARFCompileUnits(dwo_dwarf, symbol_map, &cu, sink);
+    if (!dwo_info.dwo_name.empty()) {
```
EricRahm: Super minor nit: I think we can just drop this outer `if`.

EricRahm (changes requested):
Thanks again for including tests here, just a couple very minor suggestions.

PR #424, opened 2025-10-29
Fix DWARF 5 range list handling for DW_FORM_sec_offset.

rjmansfield:
Implement handling for DW_RLE_base_address, DW_RLE_start_end and DW_RLE_start_length.

Without the change, bloaty crashes on some DWARF 5 binaries with "Overflow in vm range".

### DIFF
--- a/src/dwarf.cc
+++ b/src/dwarf.cc
@@ -455,7 +455,7 @@ void AddDIE(const dwarf::CU& cu, const GeneralDIE& die,
   }
 
   // Sometimes a location is given as an offset into debug_loc.
-  if (die.location_uint64) {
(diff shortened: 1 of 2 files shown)


EricRahm:
@rjmansfield Thanks this looks like a nice improvement! I took a cursory look at this, but will need a little more time to get up to speed on the dwarf5 specifics, overall it makes sense to me. Ideally we'd include a test for this as well.

rjmansfield:
Sounds good.  I hit this issue with on a more complicated binary and it was challenging to distill a test case from it.  I'll take another crack at coming up with a test for it though.

PR #425, opened 2025-10-29
Improve dSYM loading for Mach-O binaries

rjmansfield:
Adds two features for better dSYM support on macOS:

1. Add Mach-O specific --dsym option
   - Accepts both flat dSYM companion files and dSYM bundles
   - Handles multiple dSYM files for multiple input binaries

2. Automatic dSYM discovery
   - Searches for <binary>.dSYM/Contents/Resources/DWARF/<binary>
   - Only loads dSYM if build IDs match to prevent mismatches

This improves the macOS user experience by eliminating the need to manually specify the binary file inside of the dSYM bundle.

### DIFF
--- a/doc/using.md
+++ b/doc/using.md
@@ -327,6 +327,28 @@ can create with `dsymutil`:
 ```
 $ dsymutil bloaty
 $ strip bloaty  (optional)
+$ ./bloaty -d symbols bloaty  # Auto-discovers bloaty.dSYM
+```
+
+Bloaty will automatically discover and load debug symbols from
+`bloaty.dSYM/Contents/Resources/DWARF/bloaty` when analyzing `bloaty`.
+
+If you used `dsymutil -o` to write the dSYM to a different directory,
+you can explicitly specify its location:
+
+```
+$ dsymutil bloaty -o /path/to/symbols/bloaty.dSYM
+$ ./bloaty -d symbols --dsym=/path/to/symbols/bloaty.dSYM bloaty
+```
+
+**Note:** The `--dsym` option is also useful for debugging issues with auto-loading dSYMs.
+When you explicitly specify `--dsym`, Bloaty will throw informative error messages
+if the dSYM file is invalid or has a mismatched build ID, rather than silently
+ignoring the file as it does during auto-discovery.
+
+For cross-platform compatibility, `--debug-file` still works:
+
+```
 $ ./bloaty -d symbols --debug-file=bloaty.dSYM/Contents/Resources/DWARF/bloaty bloaty
 ```
 
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -29,6 +29,7 @@ typedef size_t z_size_t;
 
 #include <atomic>
 #include <cmath>
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <limits>
@@ -1526,6 +1527,9 @@ class Bloaty {
   void AddFilename(const std::string& filename, bool base_file);
   void AddDebugFilename(const std::string& filename);
   void AddSourceMapFilename(const std::string& filename);
+  void AddDsymFilename(const Options& options, const std::string& filename);
+  void TryAutoLoadDsym(const std::string& binary_path,
+                       const std::string& build_id);
 
   size_t GetSourceCount() const { return sources_.size(); }
 
@@ -1572,6 +1576,15 @@ class Bloaty {
                          std::vector<std::string>* out_build_ids) const;
 
   std::unique_ptr<ObjectFile> GetObjectFile(const std::string& filename) const;
+
+  bool TryGetBuildId(const std::string& filename, std::string* build_id);
+  bool TryValidateAndAddDsym(const std::string& dsym_path,
+                             const std::string& main_binary_path,
+                             bool strict_validation);
+
+  std::filesystem::path ConstructDsymPath(const std::filesystem::path& binary_path);
+  std::filesystem::path GetDsymInternalPath(const std::filesystem::path& dsym_bundle,
+                                            const std::string& binary_name);
 
   const InputFileFactory& file_factory_;
   const Options options_;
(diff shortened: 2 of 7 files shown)


rjmansfield:
First patch failed on Windows with the implicit conversion:

`D:\a\bloaty\bloaty\bloaty\src\bloaty.cc(1637,9): error C2664: 'void bloaty::Bloaty::AddDebugFilename(const std::string &)': cannot convert argument 1 from 'std::filesystem::path' to 'const std::string &' [D:\a\bloaty\bloaty\build\bloaty\libbloaty.vcxproj]`

Updated patch should resolve that issue.

src/bloaty.cc
```diff
@@ -1607,6 +1613,96 @@ void Bloaty::AddSourceMapFilename(const std::string& filename) {
...
+  std::filesystem::path dsym_bundle = binary_dir / (binary_name + ".dSYM");
+  std::filesystem::path dsym_file = dsym_bundle / "Contents" / "Resources" / "DWARF" / binary_name;
+
+  if (verbose_level > 0) {
+    fprintf(stderr, "Looking for dSYM at: %s\n", dsym_file.string().c_str());
+  }
+
+  if (std::filesystem::exists(dsym_file)) {
+    // Verify that the dSYM file has the same build ID as the main binary
+    try {
```
EricRahm: What part of this would actually throw an exception? I wonder if we can avoid that.

EricRahm (changes requested):
@rjmansfield Took a quick look and decided I need to brush up on `dSYM` before going too deep, so consider this a very limited first pass.

I made some super minor comments around logging -- my general feedback is maybe remove some of it, particularly for expected failures where we're speculatively looking for debug info.

src/bloaty.cc
```diff
@@ -1607,6 +1613,96 @@ void Bloaty::AddSourceMapFilename(const std::string& filename) {
...
+  std::filesystem::path binary_file(binary_path);
+  std::string binary_name = binary_file.filename().string();
+  std::filesystem::path binary_dir = binary_file.parent_path();
+
+  // Look for dSYM bundle: <binary>.dSYM/Contents/Resources/DWARF/<binary>
+  std::filesystem::path dsym_bundle = binary_dir / (binary_name + ".dSYM");
+  std::filesystem::path dsym_file = dsym_bundle / "Contents" / "Resources" / "DWARF" / binary_name;
+
+  if (verbose_level > 0) {
+    fprintf(stderr, "Looking for dSYM at: %s\n", dsym_file.string().c_str());
```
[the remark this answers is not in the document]
rjmansfield: OK, sounds good. I've trimmed most of the logging.

src/bloaty.cc
rjmansfield: Well GetObjectFile() may throw if the dsym binary is invalid and we're speculating trying to load the dSYM. I didn't think it should fail in that case. I've added a comment here.

EricRahm:
Just a heads up, I'll be away for a week or so but should be able to take a look next week!

EricRahm:
@haberman mind taking a look, this is improving dsym support. It's adding a new flag `--dsym`

doc/using.md
```diff
@@ -327,6 +327,20 @@ can create with `dsymutil`:
...
 $ strip bloaty  (optional)
+$ ./bloaty -d symbols bloaty  # Auto-discovers bloaty.dSYM
+```
+
+Bloaty will automatically discover and load debug symbols from
+`bloaty.dSYM/Contents/Resources/DWARF/bloaty` when analyzing `bloaty`.
+You can also explicitly specify a dSYM bundle:
+
+```
+$ ./bloaty -d symbols --dsym=bloaty.dSYM bloaty
```
haberman: This example is a little confusing because, if I'm understanding right, it tells Bloaty to load a bundle that it would have auto-loaded even without the flag.

Should this example use a different dSYM filename? Would the main use case be loading a dsym from a different directory?

src/bloaty.cc
```diff
@@ -1663,6 +1670,78 @@ void Bloaty::AddSourceMapFilename(const std::string& filename) {
...
+      WARN("dSYM bundle '$0' contains no debug files matching input binaries",
+           filename);
+    }
+  } else {
+    THROWF("'$0' is not a dSYM bundle directory", filename);
+  }
+}
+
+// Automatically discover and load dSYM file for a Mach-O binary
+void Bloaty::TryAutoLoadDsym(const std::string& binary_path,
```
haberman: From the user docs, I would expect that auto-loading of a dSYM performs exactly the same logic as explicitly specifying `--dsym`, but that doesn't seem to be the case.

Is there a reason to have two distinct functions for this? If so, can we add more explanation in the user docs?

doc/using.md
rjmansfield: OK, I will change it to a non-autoloaded path in the example and further explain. 

Ideally the high runner case is that user will use clang as linker driver with` -g` and then clang will spawn `dsymutil` and write the dSYM in the same directory as the binary and then debug information will be seamlessly loaded. The explicit option `--dsym` option would be only for when `dsymutil` gets invoked directly and written to different directory than the executable where autoloading doesn't find them.

src/bloaty.cc
rjmansfield: They have slightly different semantics and have different error handling.

For auto loading, I think lookup should be silent best effort but not throw/error if it fails. Where if a user specifies explicitly a `--dsym` and it's invalid or missing, that it should be reported to the user e.g. AddDsymFilename will throw if if file doesn't exist or if not a dSYM bundle directory.

I can explain in the documentation, but I think there's probably a little bit of refactoring that I can do here.

rjmansfield: @haberman @EricRahm Would you mind having another look at the current patch when you have a chance? Thanks.

haberman (approved):

EricRahm:
@haberman I think we've resolved all the feedback and our tests are passing, should we go ahead and land this?

haberman:
Yep SGTM (I approved the changes above).

PR #429, opened 2025-10-30
Replace libcxxabi demangler with demumble's LLVM and Swift demanglers

### LINKED ISSUE #110: Rust symbol demangling

### LINKED ISSUE #394: demangling failures
I was running `bloaty libm.a -n 0 --demangle=full -d symbol` and noticed some functions aren't getting demangled, like:
```
_ZNK22__llvm_libc_21_0_0_git6fputil11DyadicFloatILm64EE10generic_asIDF16_Lb1EEENS_3cpp9enable_ifIXaasr3cppE19is_floating_point_vIT_Eltsr6FPBitsIS6_EE12FRACTION_LENLm64EES6_E4typeEv
```
llvm-cxxfilt was able to demangle that as:
```
__llvm_libc_21_0_0_git::cpp::enable_if<cpp::is_floating_point_v<_Float16> && FPBits<_Float16>::FRACTION_LEN < 32ul, _Float16>::type __llvm_libc_21_0_0_git::fputil::DyadicFloat<32ul>::generic_as<_Float16, true>() const
```
I'm guessing the new for C23 type `_Float16` is problematic? In that vein, I bet `_Float128` is, too?

rjmansfield:
This replaces the old libcxxabi-based C++ demangler with demanglers
 from the demumble project, which includes LLVM's demangler for C++
Itanium, Rust and Microsoft symbols, as well as the Swift demangler.

This improves Bloaty's C++ demangling support, as well as adds support
for demangling Swift, Rust and Microsoft/MSVC symbols.

Fixes #110
Fixes #394

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -60,7 +60,9 @@ typedef size_t z_size_t;
 #include "bloaty.pb.h"
 #include "google/protobuf/io/zero_copy_stream_impl.h"
 #include "google/protobuf/text_format.h"
+#include "llvm/Demangle/Demangle.h"
 #include "re.h"
+#include "swift/Demangling/Demangle.h"
 #include "util.h"
 
 using std::string_view;
@@ -162,9 +164,6 @@ static std::string CSVEscape(string_view str) {
   }
 }
 
-extern "C" char* __cxa_demangle(const char* mangled_name, char* buf, size_t* n,
-                                int* status);
-
 std::string ItaniumDemangle(string_view symbol, DataSource source) {
   if (source != DataSource::kShortSymbols &&
       source != DataSource::kFullSymbols) {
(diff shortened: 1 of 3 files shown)


rjmansfield:
I have a PR open ( https://github.com/nico/demumble/pull/40) to update the demanglers in demumble as they're about a year  out of date, and don't have the latest llvm and swift changes. However, even if Bloaty adopts demumble at 10e00fb, it will resolve the issues and be an overall improvement in demangling. Then in the future, if demumble is updated, it should be relatively straight forward to update the submodule.

src/bloaty.cc
```diff
@@ -182,15 +181,41 @@ std::string ItaniumDemangle(string_view symbol, DataSource source) {
...
+    } else if (char* ms = llvm::microsoftDemangle(
+                   demangle_from, NULL, NULL,
+                   llvm::MSDemangleFlags(llvm::MSDF_NoAccessSpecifier |
+                                         llvm::MSDF_NoCallingConvention |
+                                         llvm::MSDF_NoMemberType |
+                                         llvm::MSDF_NoReturnType))) {
+      std::string ret(ms);
+      free(ms);
+      return ret;
+    } else if (swift::Demangle::isSwiftSymbol(demangle_from)) {
```
EricRahm: Should we demangle `rust` in the short case as well? If not can we add a comment why it's skipped?

src/bloaty.cc
```diff
@@ -182,15 +181,41 @@ std::string ItaniumDemangle(string_view symbol, DataSource source) {
     if (absl::debugging_internal::Demangle(demangle_from.data(), demangled,
```
EricRahm: We're somewhat interested in moving off of absl's `Demangle`, could we replace that with demumble as well? 

This could be a follow up issue / discussion, just wanted to highlight it while we're here.

EricRahm:
> I have a PR open ( [nico/demumble#40](https://github.com/nico/demumble/pull/40)) to update the demanglers in demumble as they're about a year out of date, and don't have the latest llvm and swift changes. However, even if Bloaty adopts demumble at 10e00fb, it will resolve the issues and be an overall improvement in demangling. Then in the future, if demumble is updated, it should be relatively straight forward to update the submodule.

Landing as is and the updating the submodule makes sense to me, I'll add a note to that PR that this would be useful for Bloaty and the Fuchsia project.

src/bloaty.cc
rjmansfield: In the latest abseil-cpp, absl::debugging_internal::Demangle's handles Rust demangling. i.e. the test I added tests demangling them in the short case as well.

EricRahm: Gotcha, can we just add a comment that absl covers it for now.

PR #430, opened 2025-10-30
Add missing --raw-map and --version to help text.

### LINKED ISSUE #371: --raw-map option missing from help string
The `--raw-map` argument is missing from the [usage string](https://github.com/google/bloaty/blob/main/src/bloaty.cc#L1935-L1980) displayed when `bloaty --help` is run. Adding this command to the usage string is easy and would make the command discoverable.

rjmansfield:
Fixes #371


EricRahm (approved):
lgtm, thanks for taking care of this!

PR #442, opened 2025-11-06
Fix config file handling of max_rows_per_level: 0

### LINKED ISSUE #375: Bloaty accepts `-n 0` but not `max_rows_per_level: 0` when using `-c` option
Title is self explanatory. When bloaty is given `max_rows_per_level: 0` via `bloaty -c bloaty.textproto` file it errors with:

```
bloaty: max_rows_per_level must be at least 1
```

However, specifying `bloaty -n 0` on the command line does the right thing as expected.

rjmansfield:
The command line option -n 0 has handling to convert 0 to INT64_MAX for "unlimited", but the config file option max_rows_per_level: 0 was not doing the same.

Fixes #375


EricRahm (approved):
lgtm, thanks for making this consistent!

PR #443, opened 2025-11-06
Fix system capstone include path for misconfigured pkg-config files

### LINKED ISSUE #342: building on macOS is tricky due to capstone.h inclusion
Almost the same issue as https://github.com/raspberrypi/openocd/issues/30 - on macOS bloaty's cmake file fails to find `capstone.h` due to the fact that Homebrew stores capstone headers in a different location than bloaty's cmake file expects (see the linked issue for details)

I am just creating this issue to mention it is a problem; and also to let other macOS users know the workaround:

```
export CXXFLAGS=-isystem\ /opt/homebrew/Cellar/capstone/4.0.2/include 
cmake -B build -G Ninja -S .
cmake --build build
```

rjmansfield:
Some pkg-config files incorrectly set the include directory to <path/to/include>/capstone instead of <path/to/include>. Since bloaty uses #include "capstone/capstone.h", it expects the parent directory.

This fix detects when the pkg-config path ends with /capstone and strips it, ensuring compatibility with both correct and incorrect pkg-config files.

Fixes #342

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,12 @@ if(UNIX OR MINGW)
     endif()
   endif()
   if(CAPSTONE_FOUND)
+    # Work around pkg-config files that incorrectly point to include/capstone
+    # instead of include e.g. homebrew's capstone package
+    get_filename_component(CAPSTONE_DIR_NAME "${CAPSTONE_INCLUDE_DIRS}" NAME)
+    if(CAPSTONE_DIR_NAME STREQUAL "capstone")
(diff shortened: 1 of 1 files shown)


EricRahm (approved):
Nice clean solution, thanks for taking care of this!

EricRahm:
Our builders don't cover this, but tested locally w/ `brew` and confirmed this fixes the previously broken configure step.
```
brew install capstone
curl -L https://github.com/google/bloaty/commit/1ef2c5d07135754bbed46880ff628b61305726e4.patch | git apply -
rm -rf build/
cmake -S . -B build -G Ninja -D BLOATY_PREFER_SYSTEM_CAPSTONE=YES
cmake --build build
```

PR #444, opened 2025-11-06
Implement -d inlines support for Mach-O.

rjmansfield:


EricRahm (changes requested):
Overall lgtm, nice and straightforward. Thanks for including the test as well. Only small request is to add some details to the commit message and include the version of `clang` you used to generate the binary.

rjmansfield:
> Overall lgtm, nice and straightforward. Thanks for including the test as well. Only small request is to add some details to the commit message and include the version of `clang` you used to generate the binary.

Sounds good. FWIW I tried to make this into a yaml test instead of using a prebuilt binary but I think I've found a obj2yaml/yaml2obj bug writing out the dsym.  Since I wasn't sure how long it'll take to fix, and how long before the test could rely on updated yaml2obj, I opted for using a binary instead.   However if you want, I can look at the yaml2obj bug more and then update the test with a comment to point to the llvm bug report (if needed)

EricRahm:
>However if you want, I can look at the yaml2obj bug more and then update the test with a comment to point to the llvm bug report (if needed)

Having an upstream bug filed and referenced would be great.

rjmansfield:
> > However if you want, I can look at the yaml2obj bug more and then update the test with a comment to point to the llvm bug report (if needed)
> 
> Having an upstream bug filed and referenced would be great.

I was able to use generate a minimal YAML test and work around the upstream bug. https://github.com/llvm/llvm-project/issues/166993 for reference.

rjmansfield:
@EricRahm can you have another look at this when you have a chance?

EricRahm:
> @EricRahm can you have another look at this when you have a chance?

Thanks for the reminder! I put some time on my schedule tomorrow. I'd like to also loop in @haberman since this has potential user facing changes.

haberman:
This looks good to me. The test file is a little bigger than we generally like, but maybe that's necessary to account for the large amount of DWARF data generated for inlines.

rjmansfield:
> This looks good to me. The test file is a little bigger than we generally like, but maybe that's necessary to account for the large amount of DWARF data generated for inlines.

OK, well I was able to trim it down a bit by ~27% by generating the yaml from the inline c test with -g1 without effecting the test behaviour.

Issue #351, opened 2023-07-18
Allow builders to prefer using system libraries


max0x53:
Currently it is possible to set `BLOATY_PREFER_SYSTEM_CAPSTONE` to use the bundled capstone version. Expand this for the other libraries to enable more flexibility when building.

EricRahm:
Thanks for the report and patch @max0x53. We've recently added the ability to configure preferring the system `protobuf` and `re2`. Still lacking is `absl` and `zlib`. Things have bitrotted a bit, I'll put up a clean PR for that.

PR #352, opened 2023-07-18
Allow system libraries to be included by cmake flag

max0x53:
Closes #351

PR #446, opened 2025-11-06
Add config prefs for system absl and zlib

EricRahm:
Allow builders to select whether or not to prefer the system `absl` and `zlib`. This can be useful when there are incompatibilities between a system library and a bundled library such as `protobuf` and `absl`.

New config flags preserve existing behavior by defaulting to `YES`:
  - BLOATY_PREFER_SYSTEM_ABSL
  - BLOATY_PREFER_SYSTEM_ZLIB

Fixes: #351
Fixes: #353


EricRahm:
This is a slightly expanded and updated version of #352.

PR #458, opened 2025-11-20
Disable assertions in demumble sources

EricRahm:
Turn off assertions when building `demumble` sources by specifying
`NDEBUG`. This fixes an issue tripping up fuzzing, in this case
`demumble` already handles the issue gracefully.

Fixes https://github.com/google/bloaty/issues/457

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -289,6 +289,33 @@ include_directories("${CMAKE_CURRENT_BINARY_DIR}/src")
 include_directories(third_party/demumble/third_party/llvm/include)
 include_directories(third_party/demumble/third_party/swift/include)
 
+
+# Demumble has an open issue build as a library.
+# See https://github.com/nico/demumble/issues/39
+set(DEMUMBLE_SOURCES
+    third_party/demumble/third_party/llvm/lib/Demangle/Demangle.cpp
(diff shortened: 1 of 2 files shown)


EricRahm:
fyi @rjmansfield seems like the most pragmatic solution, but may impact demangling older swift binaries (I presume).

rjmansfield:
> fyi @rjmansfield seems like the most pragmatic solution, but may impact demangling older swift binaries (I presume).

Thanks for letting me now. Asserts are used in new demangler as well, so there might be other malformed symbols that trigger asserts for debug builds. I believe expected behaviour of swift demangle `swift demangle` is to fail gracefully in release mode,  i.e. return an unmangled symbol back if malformed. So maybe it would make sense to always define NDEBUG for the swift demangler source files.

EricRahm:
I think the fuzzer is unhappy `assert` is being triggered in a `Release` build (perhaps as opposed to something that can't be disabled with `NDEBUG`), I'm not super up to speed on the cluster fuzz behavior here. Looking at the swift code it should probably at least have an early return after the assert in case `NDEBUG` is defined, but perhaps you're right, if we define `NDEBUG` for the swift sources it may handle it gracefully.

EricRahm:
@rjmansfield Good call, explicitly setting `NDEBUG` does the trick.

rjmansfield:
@EricRahm Awesome thanks!

Issue #469, opened 2025-12-05
Ignore SHF_ALLOC sections in VM memory map


kesyog:
We've run into an issue where symbols in output sections marked `INFO` in our linker script show up in our bloaty output with any of `-d symbols`, `-d sections,symbols`, or `-d segments,symbols`.

I think the root cause is that:
* Our executable's VM memory map has valid `SHF_ALLOC` regions with low addresses near 0
* In `lld`, non-`SHF_ALLOC` sections always get 0 `sh_addr`.
* AFAICT, once bloaty has built up a range map of the memory space, it will assign symbols to a section purely based on the symbol's address. The address of our non-`SHF_ALLOC` `INFO` section overlaps with our `SHF_ALLOC` section near 0, which confuses Bloaty.

Would it be reasonable to check each symbol's section header index (`st_shndx`) and discard it from VM space if the corresponding section is not `SHF_ALLOC`?

EricRahm:
@kesyog Thanks for the report! Your analysis and proposed fix sound about right. It would be helpful if you could provide an example binary that reproduces this for you or a simplified example of sources, linker script, and build commands so that we can repro locally.

kesyog:
I threw together a small reproducer: https://github.com/kesyog/bloaty-469.

For more context, our use case is an embedded project that stores a [Pigweed tokenization database](https://pigweed.dev/pw_tokenizer/tokenization.html) as metadata in the linked executable.

EricRahm:
This is super helpful, thanks for putting together the reproducer!

Just confirming what you noted:
```
 % readelf -S build/app
There are 30 section headers, starting at offset 0x1d10:

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .text             PROGBITS         0000000000000000  00001000
       00000000000000ba  0000000000000000  AX       0     0     16
<snip>
  [25] .metadata         PROGBITS         0000000000000000  000016c0
       0000000000000024  0000000000000000   W       0     0     16
<snip>
```

So we see `.text` and `.metadata` both have an `0x0` address. From the example we're adding items to `.metadata` section via an annotation:
```c
// These metadata symbols are placed in an INFO output section
static int metadata1 __attribute__((used, section(".metadata"))) = 6;
static int metadata2 __attribute__((used, section(".metadata"))) = 5;
[...]

PR #471, opened 2025-12-09
elf: Only add SHF_ALLOC symbols to VM map

EricRahm:
Only symbols in sections marked with `SHF_ALLOC` should be added to the
VM map. This adds a check to filter non-`SHF_ALLOC` entries and fixes an
issue where info related sections could have addresses that conflicted
with `SHF_ALLOC` sections that actually affect the VM size.

Fixes: https://github.com/google/bloaty/issues/469


EricRahm:
@haberman This was the potential "simple" fix for #469. The reporter confirmed it fixed their issue locally.

haberman (approved):

Issue #345, opened 2023-06-08
Can't read index 0 from strtab


eliao:
Great tool!

Sometimes when I run it with "-d symbols" I get this error:
bloaty: can't read index 0 from strtab, total size is 365881

I am unclear why it can't read index 0.  Unfortunately, I'm not able to share my binary that is causing this.

It'll come and go with changes in the code.  I don't particularly care about my strtab at the moment; is there a way to get bloaty to not exit here?

unwiredben:
This is likely a duplicate of #217

Issue #217, opened 2020-11-12
Symbols with no name should use the address for their "name"


haberman:
Sometimes symbols are present in the symbol table, but the strtab index is 0. This indicates a symbol that has had its name removed, even though the symbol table entry is present.

Right now Bloaty will error out on these symbols with an error like:

> bloaty: can't read index 0 from strtab, total size is 254049

Instead we should use the address of the symbol as its name in this case.

PR #472, opened 2025-12-11
Allow undefined symbol names

EricRahm:
When processing symbol table names an error would be thrown if the the
name was undefined (0) resulting in failures to process files like:

> bloaty: can't read index 0 from strtab, total size is 254049

This modifies the symbol table processing to return an empty string if
the name is undefined. An empty string will trigger the existing logic
to generate a synthetic "Anonymous" name instead of crashing.

Fixes: #217
Fixes: #345


EricRahm:
@haberman pretty small fix, builds on #467 where we added support for synthetic names.

Issue #153, opened 2018-12-07
Corrupt Mach-O file


kellycampbell:
Bloaty gives a `Corrupt Mach-O file` error on a file.

`lipo -detailed_info` gives this output:

```
Fat header in: REDACTEDSdk
fat_magic 0xcafebabe
nfat_arch 5
architecture armv7
    cputype CPU_TYPE_ARM
    cpusubtype CPU_SUBTYPE_ARM_V7
    offset 108
    size 8291248
    align 2^2 (4)
architecture armv7s
    cputype CPU_TYPE_ARM
    cpusubtype CPU_SUBTYPE_ARM_V7S
    offset 8291356
    size 8307560
    align 2^2 (4)
architecture i386
    cputype CPU_TYPE_I386
    cpusubtype CPU_SUBTYPE_I386_ALL
    offset 16598916
    size 8294024
    align 2^2 (4)
architecture x86_64
    cputype CPU_TYPE_X86_64
    cpusubtype CPU_SUBTYPE_X86_64_ALL
    offset 34485416
    size 8454008
    align 2^3 (8)
architecture arm64
    cputype CPU_TYPE_ARM64
    cpusubtype CPU_SUBTYPE_ARM64_ALL
    offset 24892944
    size 9592472
    align 2^3 (8)
```

The hexdump of the first part of the file is this:

```
0000000 ca fe ba be 00 00 00 05 00 00 00 0c 00 00 00 09  >................<
0000020 00 00 00 6c 00 7e 83 b0 00 00 00 02 00 00 00 0c  >...l.~..........<
0000040 00 00 00 0b 00 7e 84 1c 00 7e c3 68 00 00 00 02  >.....~...~.h....<
0000060 00 00 00 07 00 00 00 03 00 fd 47 84 00 7e 8e 88  >..........G..~..<
0000100 00 00 00 02 01 00 00 07 00 00 00 03 02 0e 34 a8  >..............4.<
0000120 00 80 ff 78 00 00 00 03 01 00 00 0c 00 00 00 00  >...x............<
0000140 01 7b d6 10 00 92 5e 98 00 00 00 03 21 3c 61 72  >.{....^.....!<ar<
0000160 63 68 3e 0a 23 31 2f 32 30 20 20 20 20 20 20 20  >ch>.#1/20       <
0000200 20 20 20 20 31 35 31 39 32 32 31 38 35 30 20 20  >    1519221850  <
[...]

PR #480, opened 2026-01-15
Add segment-aware VM addressing for universal binary support

rjmansfield:
Bloaty currently cannot correctly analyze macho universal binaries. When processing a universal binary containing multiple architectures. e.g. arm64 and x86_64, each architecture slice should have its own virtual address space. However Bloaty's current implementation uses a single flat address space, causing these overlapping addresses to conflict and produce incorrect results. As suggested in https://github.com/google/bloaty/issues/153#issuecomment-3587223309, this adds a segment id, and then updates the logic to handle multiple address spaces. 

This bulk of the changes is introducing a VMAddr structure which contains a segment identifier and an address. The more challenging and complex changes are to ComputeRollup which required relaxing the logic for secondary maps to accommodate things like, padding or gaps (otherwise these maps trigger asserts in the previous code). With the segment infrastructure changes in place, the macho changes were fairly straightforward.

@haberman Does this approach seem reasonable to you? I hope my understanding of the rollup algorithm and the changes are correct. All of the existing tests are passing for me locally but it's possible I've missed something, so any feedback would be appreciated.

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -1149,7 +1149,7 @@ bool RangeSink::IsVerboseForVMRange(uint64_t vmaddr, uint64_t vmsize) {
     RangeMap vm_map;
     RangeMap file_map;
     bool contains = false;
-    vm_map.AddRangeWithTranslation(vmaddr, vmsize, "", translator_->vm_map,
+    vm_map.AddRangeWithTranslation(vmaddr, vmsize, "", translator_->GetVMMap(arch_index_),
                                    false, &file_map);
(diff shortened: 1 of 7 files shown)


rjmansfield:
Updated based on previous feedback, which simplifies things. Note that the segment/slice names depend on changes added in https://github.com/google/bloaty/pull/481 so they're currently stubbed out currently  I can rebase or fix if/when it lands.

rjmansfield:
Updated commit now that https://github.com/google/bloaty/pull/481 has landed

PR #481, opened 2026-01-16
Add -d archs data source for Mach-O universal binaries

rjmansfield:
Implements the 'archs' data source to break down universal binaries by architecture.

This allows users to:

- View file size breakdown by architecture: bloaty -d archs
- Filter to specific architecture: bloaty -d archs,segments --source-filter=arm64
- Hierarchical breakdown: bloaty -d archs,segments

This was proposed in https://github.com/google/bloaty/issues/153#issuecomment-3587223309 to show the breakdown per architecture using the --source-filter=<arch> option.

Note: this has the current limitation which shows incorrect VM sizes because both architectures have overlapping VM addresses, so the workaround would be --source-filter=<arch>, or to only use --domain=file. The overlapping VM addresses  requires separate address spaces are used. e.g. https://github.com/google/bloaty/pull/480

### DIFF
--- a/src/bloaty.cc
+++ b/src/bloaty.cc
@@ -85,6 +85,7 @@ struct DataSourceDefinition {
 
 constexpr DataSourceDefinition data_sources[] = {
     {DataSource::kArchiveMembers, "armembers", "the .o files in a .a file"},
+    {DataSource::kArchs, "archs", "architecture slices in universal binaries"},
     {DataSource::kCompileUnits, "compileunits",
      "source file for the .o file (translation unit). requires debug info."},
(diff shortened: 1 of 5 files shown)


haberman (approved):

rjmansfield:
Previous patch failed on non-darwin platforms because I missed  including "third_party/darwin_xnu_macho/mach/machine.h" for them. Should be fixed now.

rjmansfield:
Can the CI be kicked for the updated patch?

rjmansfield:
Looks like the CI jobs all passed so can this be merged?

PR #498, opened 2026-06-02
fix out-of-bounds read for zero-length ELF note name

sahvx655-wq:
A note with n_namesz=0 makes name_.size()-1 underflow in NoteIter::Next, so the trailing-NUL check reads name_[SIZE_MAX] out of bounds; guard it with !name_.empty().


sahvx655-wq:
gentle ping

PR #499, opened 2026-06-03
bounds-check source file index in source map vlq parser

sahvx655-wq:
Spotted this reading the source map parser. The sources array is bounds-checked nowhere before ForEachVLQSegment indexes it with sources[source_file], but source_file comes straight out of the mappings VLQ stream and is a signed int32 that can be negative or past the end of sources. A crafted .map walks off the vector. Reject the out-of-range index before the lookup, the same way the surrounding parser already throws on malformed input.


sahvx655-wq:
any update?

PR #501, opened 2026-07-20
bounds check source index before indexing sources in source maps

fr-manvi:
Bloaty reads source maps supplied with --source-map, and those are untrusted input just like the object files. In ForEachVLQSegment the source index starts as the second value of the first VLQ segment and is then accumulated from the deltas of every later segment, all of which come straight out of the mappings string, so it can be moved to any int32 value including negative ones. Nothing compares it against sources.size() before sources[source_file] runs, and since that is std::vector::operator[] the read simply goes wherever the index points, after which VlqSegment copies the resulting string_view into a std::string and follows a pointer that was never part of the vector. I hit it while feeding mutated maps at the wasm sourcemap tests: with sources holding a single entry, mappings of "AAAA,AgqjGAA,AAAA" moves the index to 100000 and bloaty dies on a bad address inside ForEachVLQSegment, while a smaller delta quietly prints a garbage file name instead. Checking the index against the vector at the point of use closes both, and throwing there keeps it consistent with how the rest of the parsers report malformed input. Test added alongside the existing wasm source map tests; it fails before the change and passes after, and the rest of the suite is unaffected.

### DIFF
--- a/src/source_map.cc
+++ b/src/source_map.cc
@@ -166,6 +166,10 @@ void ForEachVLQSegment(std::string_view* data,
 
     int new_values_count = ReadBase64VLQSegment(data, values);
     if (values_count >= 4) {
+      if (source_file < 0 ||
+          static_cast<size_t>(source_file) >= sources.size()) {
+        THROWF("source index $0 out of range in source map", source_file);
+      }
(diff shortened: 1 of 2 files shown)


fr-manvi:
any update?

PR #502, opened 2026-07-23
avoid out-of-bounds read on empty ELF note name

fr-manvi:
GetBuildId walks the notes in every SHT_NOTE section, and each note's name comes from NoteIter::Next as name_ = StrictSubstr(remaining_, 0, note.n_namesz). The n_namesz field is read straight out of the note header, so it is attacker-controlled just like the rest of the object. When a note carries n_namesz == 0 the name_ view is empty, and the very next line checks name_[name_.size() - 1] to trim a trailing NUL, so name_.size() - 1 wraps to SIZE_MAX and the operator[] reads outside the view. I hit it feeding a crafted ET_DYN with a single empty-name note; a build with libc++ hardening traps in NoteIter::Next, and it is on the default path since GetBuildId runs for every non-object input. Guarding the trim with !name_.empty() closes it while leaving the normal case (a name ending in NUL) unchanged. Added a yaml2obj regression alongside the other elf tests.

### DIFF
--- a/src/elf.cc
+++ b/src/elf.cc
@@ -443,7 +443,7 @@ void ElfFile::NoteIter::Next() {
   name_ = StrictSubstr(remaining_, 0, note.n_namesz);
 
   // Size might include NULL terminator.
-  if (name_[name_.size() - 1] == 0) {
+  if (!name_.empty() && name_[name_.size() - 1] == 0) {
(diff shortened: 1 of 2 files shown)


fr-manvi:
any update?

fr-manvi:
any update?

PR #505, opened 2026-08-27
tests: Ensure malformed symbol remains invalid

jriv01:
Newer versions of the Swift demangler return true on `isSwiftSymbol("_TMXfZD")` due to `_T` being a real legacy prefix for Swift mangling. This change updates the symbol name to ensure that it is invalid, and is not accidentally parsed by the Swift demangler.


haberman (approved):

jriv01:
Failed checks for Windows CI seem to be due to an issue with using windows-latest as the runner, instead of an issue with this pull request.

Opened #506 to follow up

PR #508, opened 2026-09-01
Add fallback macros for missing Mach-O constants

jriv01:
Fixes compilation failures on platforms that lack some Mach-O definitions:

```
src/macho.cc:79:30: error: use of undeclared identifier 'CPU_SUBTYPE_ARM64E'
  return subtype_proper == CPU_SUBTYPE_ARM64E;
                           ^
src/macho.cc:483:10: error: use of undeclared identifier 'LC_DYLD_CHAINED_FIXUPS'
    case LC_DYLD_CHAINED_FIXUPS:
         ^
src/macho.cc:486:10: error: use of undeclared identifier 'LC_DYLD_EXPORTS_TRIE'
    case LC_DYLD_EXPORTS_TRIE:
         ^
3 errors generated.
```

This change adds fallbacks at the top of `macho.cc` to ensure Bloaty can independently compile.


haberman:
I thought we were using all of the headers here instead of platform headers: https://github.com/google/bloaty/tree/main/third_party/darwin_xnu_macho

Do we just need to update the versions of those headers for the new definitions?

jriv01:
> Do we just need to update the versions of those headers for the new definitions?

It seems you're right, updating the versions of those headers downstream got tests passing again. In that case, we can try making the fix downstream instead of injecting fallback definitions here.

PR #475, opened 2026-01-05
dwarf: Fix SEGV when accessing uninitialized skeleton CU

EricRahm:
When processing malformed DWARF data, specifically a compile unit
where a top-level DIE attribute requires indirect address resolution
(e.g., DW_AT_GNU_dwo_id using DW_FORM_addrx), Bloaty would crash.
This occurred because `ReadTopLevelDIE` was called before `skeleton_`
was initialized in `CU::ReadHeader`, leading to a null pointer
dereference in `ReadIndirectAddress`.

This patch moves the initialization of `skeleton_` to the beginning
of `CU::ReadHeader` to ensure it is available when reading top-level
attributes.

Fixes #402
Fixes #409
Fixes #463

PR #512, opened 2026-09-09
Fix use-of-uninitialized-value when parsing GNU_dwo_id

jriv01:
`GNU_dwo_id` uses `DW_FORM_data8`, which Bloaty natively parses as a string. The parser previously checked `IsUint()`, bypassing the assignment and leaving `dwo_id_` uninitialized. This updates the parser to correctly expect a string.

Additionally updates `dw_form_implicit_const.test`. Removing this `IsUint()` check fixes Bloaty's last unsafe evaluation of `DW_FORM_implicit_const`, so the test now expects a successful run rather than a thrown exception.

### DIFF
--- a/src/dwarf/debug_info.cc
+++ b/src/dwarf/debug_info.cc
@@ -230,8 +230,11 @@ void CU::ReadTopLevelDIE(InfoReader& reader) {
             }
             break;
           case DW_AT_GNU_dwo_id:
-            if (value.IsUint()) {
-              dwo_id_ = value.GetUint(*this);
+            if (value.IsString()) {
+              string_view str = value.GetString(*this);
+              if (str.size() == 8) {
+                  dwo_id_ = ReadFixed<uint64_t>(&str);
+              }
             }
             break;
         }
--- a/src/dwarf/debug_info.cc
+++ b/src/dwarf/debug_info.cc
@@ -230,7 +230,9 @@ void CU::ReadTopLevelDIE(InfoReader& reader) {
             }
             break;
           case DW_AT_GNU_dwo_id:
-            if (value.IsString()) {
+            if (value.IsUint()) {
+              dwo_id_ = value.GetUint(*this);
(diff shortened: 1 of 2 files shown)


src/dwarf/debug_info.cc
```diff
@@ -230,8 +230,11 @@ void CU::ReadTopLevelDIE(InfoReader& reader) {
             }
             break;
           case DW_AT_GNU_dwo_id:
-            if (value.IsUint()) {
-              dwo_id_ = value.GetUint(*this);
+            if (value.IsString()) {
```
haberman: This PR removes the handling of the case where the dwo_id is a uint, but it seems to be treated as a uint by other software, for example:

https://github.com/staticfloat/rr/blob/c512416f2f186a68fade557e2a9cf1cd4fb9a2ae/src/SourcesCommand.cc#L201

https://github.com/StackDoubleFlow/LLVMStackPU/blob/70cf80a50a5d2758d5711762abe938bd67dc8fb4/llvm/tools/llvm-dwp/llvm-dwp.cpp#L200

jriv01: Added back the `value.IsUint()` check as a possible case

src/dwarf/debug_info.cc
```diff
@@ -232,6 +232,11 @@ void CU::ReadTopLevelDIE(InfoReader& reader) {
           case DW_AT_GNU_dwo_id:
             if (value.IsUint()) {
               dwo_id_ = value.GetUint(*this);
+            } else if (value.IsString()) {
```
haberman, line 235: What bug does this fix? Do we need test coverage for it?

jriv01: This fixed a bug reported internally after Fuzz testing about a year ago, but since then it actually seems like another commit fixed the same bug (#475).

We can drop this patch and stick with the fix that's already been made. Worst case we can revisit it if the need ever comes up again.

PR #514, opened 2026-09-09
tests: Anchor non-alloc-symbols test EOF

jriv01:
Without `# CHECK: FILE SIZE` to anchor the `VM MAP` block, FileCheck keeps reading into the rest of Bloaty's output. The excluded `.info` symbol correctly appears in the final summary table, triggering a false-positive `CHECK-NOT` failure.


haberman (approved):

jriv01:
This actually seems to be an issue with internal Bloaty not respecting `--raw-map`. Will look into it, but this PR can likely be closed in favor of an internal fix.

PR #516, opened 2026-09-19
Fix filter reporting

SamB:
- **Revive dead filter-reporting code in RollupOutput::PrettyPrint()**
  Clang told me that `filtered` was set but not used and I noticed that
  the subsequent conditions were testing variables which are always
  zero that have suspiciously similar names.
  

- **Disambiguate filter-reporting output**
  If we are supposed to be showing both file & vm sizes, but only one
  got filtered, be explicit about it.


SamB:
Huh. I had expected that `gh pr create -B main --fill-verbose` would still ask me to confirm things ...

